PR #27, opened 2021-09-02
Fixes #26 Incorrect signature type is used for P256KeyPair

### LINKED ISSUE #26: Incorrect signature type is used for P256KeyPair
P256KeyPair allows use `SignatureType::ES256K` for signature and disallows `SignatureType::ES256` that looks wrong. See

https://github.com/hyperledger/aries-askar/blob/494133c08eeb08423cd2b5aa03ba398a7197d884/askar-crypto/src/alg/p256.rs#L197

vimmerru:
* Fixes #26 Incorrect signature type is used for P256KeyPair
* Use compatible signature type inside of KeySign::write_signature for P256KeyPair


andrewwhitehead (approved):
Thanks!

PR #29, opened 2021-09-15
Initial BBS+ signature support crate

andrewwhitehead:
Details will assuredly change before a release, as the spec (2022 version) is still in development. This is not compatible with the BBS+ 2020 spec.


andrewwhitehead:
WIP @tplooker @mikelodder7

Issue #51, opened 2022-06-05
Transaction commit


dbluhm:
I had a question come up as I've been working with the Askar Python wrapper.

Consider these lines from the demo script:

```python
    async with store.transaction() as txn:
        # ^ should be faster within a transaction
        for idx in range(PERF_ROWS):
            await txn.insert(
                "txn",
                f"name-{idx}",
                b"value",
                {"~plaintag": "a", "enctag": "b"},
            )
        await txn.commit()
```

My question pertains to the `await txn.commit()` on the final line. As I understand it, after `commit` has been called, the handle is closed. In other words, it is impossible to do something like:

```python
    async with store.transaction() as txn:
        # ^ should be faster within a transaction
        for idx in range(PERF_ROWS):
            await txn.insert(
                "txn",
                f"name-{idx}",
                b"value",
                {"~plaintag": "a", "enctag": "b"},
            )
        await txn.commit()
        await txn.insert(
            "txn",
            f"name-one-more",
            b"value",
            {"~plaintag": "a", "enctag": "b"},
        )
        await txn.commit()
```

Therefore, calling `commit` should essentially terminate every `async with store.transaction()` block and can occur only once within the block. My question: why not call `commit` on exit of the `with` block (within the context manager itself)? This seems like an ideal case for using this pattern, in fact, from my perspective at least.

Curious to hear your rationale. Thanks!

PR #52, opened 2022-06-05
feat: auto commit txn on aexit

dbluhm:
This is an attempt to visualize what #51 would look like. I think it's this simple?

As implemented, instead of an effective "rollback by default", it will rollback if the transaction exits with an exception but otherwise commit by default. The user may still choose to manually call `rollback` before the context is exited.

### DIFF
--- a/wrappers/python/aries_askar/store.py
+++ b/wrappers/python/aries_askar/store.py
@@ -626,4 +626,11 @@ class OpenSession:
     async def __aexit__(self, exc_type, exc, tb):
         session = self._session
         self._session = None
+
+        if session.is_transaction and session.handle:
+            if exc:
+                await session.rollback()
+            else:
+                await session.commit()
+
         await session.close()


wrappers/python/aries_askar/store.py
```diff
@@ -626,4 +626,11 @@ async def __aenter__(self) -> Session:
     async def __aexit__(self, exc_type, exc, tb):
         session = self._session
         self._session = None
+
+        if session.is_transaction and session.handle:
+            if exc:
+                await session.rollback()
+            else:
+                await session.commit()
+
```
andrewwhitehead, line 635: I think the `close()` should be in an else block, as commit and rollback both close the session.

andrewwhitehead:
I think this makes sense, but I should put out another patch release before making this kind of breaking change. Also maybe a log statement when it rolls back due to an exception?

andrewwhitehead:
Planning to add this (or the equivalent) to 0.3.x shortly.

andrewwhitehead:
Added support for an autocommit flag in #162

PR #57, opened 2022-07-14
feat: add include to index so some operations don't need to access th…

reflectivedevelopment:
We found that this change improves performance when there are a large number of records in the database. With a small number of records, it doesn't appear to impact performance.

Signed-off-by: Kim Ebert <kim@indicio.tech>


andrewwhitehead (approved):
Thanks!

PR #60, opened 2022-08-01
feat(js): wrapper JavaScript

blu3beri:
- used cbindgen to create a header file
- Nodejs wrapper + tests
- React Native wrapper (for now without example app)
- Build workflow for ios and android

This PR contains a lot of `@TS-ignores` and those will be fixed in the near future as a seperate PR. I mainly wanted to merge the working version and do cleanup afterwards so integration can happen faster. 

@andrewwhitehead is there any strategy for a PR like this? Its 18.7k lines with quite a bit of wrapper repitition.


blu3beri:
@andrewwhitehead I don't know why `Build Python packages (ubuntu-latest)` is failing. I only added a headerfile and two additional workflows for iOS and Android.

andrewwhitehead:
@blu3beri There's some intermittent exception that gets thrown at shutdown on Linux and causes a failure, I'm not sure what the issue is there but it only seems to show up after the tests complete. Running the tests again usually works.

andrewwhitehead (approved):

PR #63, opened 2022-11-10
feat(wql): ported indy-wql to askar-wql

blu3beri:
Just ported the library to this location and made some idiomatic rust changes, did not change the tests so it is just plug and play.


blu3beri:
@andrewwhitehead could you rerun the tests? It seems like the same issue as before with the Python wrapper.

andrewwhitehead (approved):

PR #66, opened 2022-11-21
Update to sqlx 0.6

andrewwhitehead:


swcurran:
Is the automated test error significant?

andrewwhitehead:
It looks like the frequent failure during shutdown on Ubuntu didn't happen (although it's unpredictable, and I've only seen it happen on Github), so that's good. It also appears that an old sqlx bug where the database close method returns prior to the DB actually being closed has returned, though. I'll have to see if there's a workaround.

PR #70, opened 2022-12-16
feat(js): added expo example app

blu3beri:
dependent on: #69 

Might need to do some work hot-reloading and everything but it works with a build.

Might be good if you review this as well @TimoGlastra


wrappers/javascript/react-native-app/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "react-native-app",
+  "version": "1.0.0",
+  "scripts": {
+    "start": "expo start --dev-client",
+    "android": "expo run:android",
+    "ios": "expo run:ios",
+    "build": "exit 0",
```
TimoGlastra, line 8: I usually do "echo 'no build needed for react-native-app'" or something similar

wrappers/javascript/react-native/package.json
```diff
@@ -18,13 +18,11 @@
   "files": [
```
TimoGlastra, line 18: We should add the `native` directory here? Or is that part of another pr? As you alraedy updated the paths in the cmake / build.gradle files

TimoGlastra (approved):

wrappers/javascript/react-native/package.json
blu3beri: Did we not want to fetch them post install?

wrappers/javascript/react-native-app/package.json
blu3beri: Yeah I did `node -e console.log...` before (Windows does not have echo I believe), but that did not work. if we can do echo for windows I am okay with that.

TimoGlastra: ah yeah not sure tbh. Let's stick with this

wrappers/javascript/react-native/package.json
TimoGlastra: You're totally right 🤦

wrappers/javascript/tsconfig.json
```diff
@@ -2,11 +2,6 @@
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
     "baseUrl": ".",
-    "paths": {
-      "aries-askar-shared": ["shared/src"],
-      "aries-askar-react-native": ["react-native/src"],
-      "aries-askar-nodejs": ["nodejs/src"]
-    },
```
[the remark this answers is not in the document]
blu3beri: Yeah I was a bit confused with the tsconfigs and which did what. You are correct (just checked) that this is required for the editor.

TimoGlastra:
Python CI is suddently failing?

TimoGlastra (approved):

blu3beri:
Yeah @andrewwhitehead is this the flaky Python test failing or is there something else going on?

andrewwhitehead:
There's a semi random test failure on Ubuntu just as it shuts everything down. It seems like it might be fixed in sqlx 0.6 but that has a different issue on Windows (not releasing the database file on close).

blu3beri:
Closing this as the example app is not stable and can be added later.

PR #73, opened 2023-01-18
fix(js): Scan with more than a single result

genaris:
As discussed with @blu3beri lately, current implementation of Scan only works when there is a single result. This quick-and-dirty fix allows to retrieve all matching records from the query.

Note: To avoid further file name conflicts, this PR is based on https://github.com/hyperledger/aries-askar/pull/71, so it can be easily merged once that one is ready. Relevant file changes for this PR are those of commit https://github.com/hyperledger/aries-askar/commit/d08b728bd554316c966422317941593ee5a894bb


TimoGlastra:
> This quick-and-dirty fix allows to retrieve all matching records from the query.

What's quick and dirty about it, and do we have a plan to make it clean?

genaris:
> > This quick-and-dirty fix allows to retrieve all matching records from the query.
> 
> What's quick and dirty about it, and do we have a plan to make it clean?

It's a bit Q&D in the sense that it fixes the problem on queries but does not address other flaws on the file like the eslint errors. But I think we can review and remove eslint ignores in a further PR (not only for this file but the entire wrapper).

TimoGlastra (approved):

PR #75, opened 2023-01-19
Exposed Argon enum in Rust to use it for key derivation

Artemkaaas:
Askar exposes `StoreKeyMethod` enum (it is used for wallet creation and opening) but for using `DeriveKey` variant we need to pass `KdfMethod` object which is crate private.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -48,7 +48,11 @@ mod ffi;
 pub mod kms;
 
 mod protect;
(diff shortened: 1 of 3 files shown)


Artemkaaas:
@andrewwhitehead @WadeBarnes  Could take a look?

andrewwhitehead (approved):

PR #79, opened 2023-01-20
Change behaviour of clause negation in tag filters

### LINKED ISSUE #74: 'Not' statements in WQL queries
When working on the integration of Askar with Aries Framework JavaScript, I noticed that Askar behaves slightly different from Indy when managing WQL queries.

In the [test I'm writing](https://github.com/genaris/aries-framework-javascript/blob/d6f348b33b2a57fb9acb6ccea141ff1fc097b405/packages/askar/src/storage/__tests__/AskarStorageService.test.ts#L242), we are inserting 3 records with the following tags:
- { myTag: 'foo' }
- { anotherTag: 'bar' }
- { myTag: 'notfoobar' }

The query is for **$not: { myTag: 'notfoobar' }**, which expects to return the first two records. However, it only returns the first one. Experimenting a little bit more, I found that it only considers those records that do have a myTag: those who do not include any myTag are ignored in the query. In other words, my test interprets the query as _"Give me all the records that are not tagged as myTag = 'notfoobar'"_, while Askar interprets it as _"Give me all the records tagged with myTag that are not equal to 'foobar'"_.

Not sure if it's something related at Rust or JS-wrapper level, but before investigating further I would like to know if this behaviour is incorrect or my test (based on Indy behaviour) should be rewritten.

andrewwhitehead:
Fixes #74


TimoGlastra (approved):

swcurran:
It would be good to have a bit more detail about this change for the ChangeLog.  Or is it just a bug fix?

PR #80, opened 2023-01-21
chore(js): add @hyperledger prefix

TimoGlastra:
Adds the @hyperledger prefix to all js packages as [aries-askar-shared](https://www.npmjs.com/package/aries-askar-shared) is already taken.

It also updates the registartion of the native implementation to be automatic when you import the nodejs / react native as is done with indy-vdr.

It also sets the version in the package json to 0.1.0-dev.1 so we can make a first release.

@blu3beri is the React Native also ready for release? I think we still need the ios / android publishing to the release right? As the version is `–dev.x` and it's already integrated sort into AFJ I say we start releasing it under the official name.


wrappers/javascript/aries-askar-shared/README.md
```diff
@@ -0,0 +1,37 @@
...
+An alterative approach is to first try to require the Node.JS package, and otherwise require the React Native package:
+
+```typescript
+try {
+  require('@hyperledger/aries-askar-nodejs')
+} catch (error) {
+  try {
+    require('@hyperledger/aries-askar-react-native')
+  } catch (error) {
+    throw new Error('Could not load Indy VDR bindings')
```
genaris, line 30:
```suggestion
    throw new Error('Could not load Aries Askar bindings')
```

wrappers/javascript/aries-askar-react-native/src/index.ts
```diff
@@ -1,5 +1,11 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
```
genaris, line 1: Oh these eslint-disable lines came back! :stuck_out_tongue:

wrappers/javascript/aries-askar-nodejs/src/index.ts
```diff
@@ -1 +1,8 @@
-export { NodeJSAriesAskar } from './NodeJSAriesAskar'
+import { registerAriesAskar } from '@hyperledger/aries-askar-shared'
+
+import { NodeJSAriesAskar } from './NodeJSAriesAskar'
+
+export const ariesAskarNodeJS = new NodeJSAriesAskar()
+registerAriesAskar({ askar: ariesAskarNodeJS })
```
genaris, line 6: This method of registering based on a simple import before everything else remembers me the `reflect-metadata` that we need to import before using class-transformer. Not sure if it's completely intuitive, but certainly this method or registering is simpler than the previous one, and as long as it's properly documented (as I've seen in the README) it's fine to me.

genaris (dismissed):
Just a minor correction on a reference to Indy VDR, but LGTM!

blu3beri:
> @blu3beri is the React Native also ready for release? I think we still need the ios / android publishing to the release right? 

The bindings should be good to go. We do indeed still need the iOS and android lib. 

> As the version is `–dev.x` and it's already integrated sort into AFJ I say we start releasing it under the official name.

Yes, that would be good.

TimoGlastra:
Okay, so should we keep the react native package private then for now?

blu3beri:
> Okay, so should we keep the react native package private then for now?

Yeah I think that that would be the cleanest way. I don't think that releasing a broken version will cause issues, but setting it to private would be a good solution.

TimoGlastra:
Ok, have set react native to private, removed the eslint disables.

I think we're ready to do the release now. There's been a lot of releases to main since the last release, so ideally we create a new binary release as well. 

@andrewwhitehead is it okay for you to create a new release? If so we need to make sure to update the binary version in the `remote_path` field in `wrappers/javascript/aries-askar-nodejs/package.json` and set it to the github release tag beforehand.  Will it be `0.2.8` or `0.2.8-dev.1`?

@blu3beri does it also look good to you?

blu3beri (approved):

blu3beri:
> Ok, have set react native to private, removed the eslint disables.
> 
> I think we're ready to do the release now. There's been a lot of releases to main since the last release, so ideally we create a new binary release as well.
> 
> @andrewwhitehead is it okay for you to create a new release? If so we need to make sure to update the binary version in the `remote_path` field in `wrappers/javascript/aries-askar-nodejs/package.json` and set it to the github release tag beforehand. Will it be `0.2.8` or `0.2.8-dev.1`?
> 
> @blu3beri does it also look good to you?

Yes! All looks good to me.

blu3beri (approved):

genaris (approved):

PR #84, opened 2023-01-23
Update deprecated github actions

andrewwhitehead:
Also updates to build with Rust 1.58.0

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,39 +31,28 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
-          toolchain: 1.56
-          override: true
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -127,17 +107,15 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@master
```
TimoGlastra, line 113: using master may break the action at any time. Does the repo have a tag that we can use instead?

TimoGlastra (approved):

.github/workflows/build.yml
andrewwhitehead: It doesn't, this action uses the tag as a shorthand to select the toolchain version.

andrewwhitehead:
I'm doing some additional testing to see if I can resolve or work around the intermittent failures with the Python package testing on Ubuntu. It seems like the switch to rust 1.58 is making it more frequent.

TimoGlastra (approved):

PR #89, opened 2023-01-26
Update dependencies

andrewwhitehead:
- Updates Rust crate dependencies
- Updates to Rust 2021 edition
- Fixes clippy warnings

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,13 +5,14 @@ members = ["askar-bbs", "askar-crypto"]
 name = "aries-askar"
 version = "0.2.8-dev.3"
 authors = ["Hyperledger Aries Contributors <aries@lists.hyperledger.org>"]
-edition = "2018"
+edition = "2021"
 description = "Hyperledger Aries Askar secure storage"
 license = "MIT OR Apache-2.0"
 readme = "README.md"
 repository = "https://github.com/hyperledger/aries-askar/"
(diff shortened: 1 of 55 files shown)


Cargo.toml
```diff
@@ -71,14 +73,15 @@ path = "./askar-crypto"
...
+version = "0.6.2"
 default-features = false
 features = ["chrono", "runtime-tokio-rustls"]
 optional = true
 
 [profile.release]
-lto = true
 codegen-units = 1
+lto = true
+panic = "abort"
```
blu3beri, line 84: What's the reason behind abort?

andrewwhitehead: It removes the panic unwinding, because a panic can't cross the FFI boundary and it potentially leaves things in an invalid state. This makes the binary smaller as well.

TimoGlastra (dismissed):

andrewwhitehead:
I was planning to merge this for 0.2.8, but since it changes the Rust requirements (edition and version) it should likely be held off until 0.3.

blu3beri (dismissed):

andrewwhitehead:
Updating this PR because, based on the number of related projects creating a breakages in the current build pipeline (including `rayon` at the moment), it is not unusual at all to upgrade the Rust version dependency in a patch release.

TimoGlastra (approved):

PR #90, opened 2023-01-31
Refactor storage into a separate crate

andrewwhitehead:
Also improves the public API of 'Store' and tidies up the crate exports.

Includes updates to scan records without a category filter (preparation for profile export/migration support).

In draft due to breaking changes, awaiting another minor release.

Follow up from #89


askar-storage/Cargo.toml
```diff
@@ -0,0 +1,69 @@
...
+
+[package.metadata.docs.rs]
+features = ["all_backends"]
+no-default-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
+[features]
+default = ["all_backends", "log"]
+all_backends = ["any", "postgres", "sqlite"]
+any = []
```
blu3beri, line 22: What does the any feature do? Is it just for the `AnyBackend`?

askar-storage/src/options.rs
```diff
@@ -126,7 +136,9 @@ fn percent_encode_into(result: &mut String, s: &str) {
     push_iter_str(result, utf8_percent_encode(s, NON_ALPHANUMERIC))
 }
 
+/// A trait implemented by types that can be converted into Options
 pub trait IntoOptions<'a> {
```
blu3beri, line 140: Is there a reason why `TryInto` for Options is not used? If this for some lower Rust version compatibility?

askar-storage/tests/backends.rs
```diff
@@ -5,228 +5,117 @@ mod utils;
...
-                db.close().await.expect(ERR_CLOSE);
-            })
+            $run(super::utils::db_txn_contention)
         }
     };
 }
 
 fn log_init() {
-    env_logger::builder().is_test(true).try_init().unwrap_or(());
+    // env_logger::builder().is_test(true).try_init().unwrap_or(());
```
blu3beri, line 109: Can we remove this function or enable it again for test logging?

src/ffi/error.rs
```diff
@@ -60,11 +60,17 @@ pub extern "C" fn askar_get_current_error(error_json_p: *mut *const c_char) -> E
...
+        code: usize,
+        message: String,
+    }
+
     if let Some(err) = Option::take(&mut *LAST_ERROR.write().unwrap()) {
         let message = err.to_string();
         let code = ErrorCode::from(err.kind()) as usize;
         // let extra = err.extra();
-        json!({"code": code, "message": message}).to_string()
+        serde_json::to_string(&ErrorJson { code, message }).unwrap()
```
blu3beri, line 73: Could we unwrap here to a custom error with code `100` or something? Would avoid a panic and allow the library to handle the error here.

blu3beri (approved):
This is looking great! I see some minor things changed in the python wrapper, are there any breaking changes from an FFI perspective, i.e. the JavaScript wrapper must also be updated? If so, I am happy to pick that up.

andrewwhitehead:
> This is looking great! I see some minor things changed in the python wrapper, are there any breaking changes from an FFI perspective, i.e. the JavaScript wrapper must also be updated? If so, I am happy to pick that up.

Most of the changes there are added docstrings, but the `category` filter is now optional (can be null) for `scan_start`, `count`, `fetch_all` and `remove_all`.

src/ffi/error.rs
andrewwhitehead: I don't think there's any reason why it would panic under normal circumstances. I believe we could also construct the serde_json types directly instead of using a wrapper type.

askar-storage/src/options.rs
andrewwhitehead: It's actually because that creates issues accepting a straight Options instance - it implements `From<Self>` which means `TryFrom<Self>` is implemented with the `Infallable` error type, not the crate's error type.

PR #93, opened 2023-02-02
chore: update version to 0.2.8-dev.1

TimoGlastra:
Signed-off-by: Timo Glastra <timo@animo.id>


blu3beri (approved):
If we do not have to update the versions in the `package.json`s then it can be merged!

TimoGlastra:
> If we do not have to update the versions in the `package.json`s then it can be merged!

Those are already set to `0.1.0-dev.1` (this will be the first version)

genaris (approved):

PR #94, opened 2023-02-03
feat(js): setup React Native

blu3beri:

### DIFF
--- a/build-xcframework.sh
+++ b/build-xcframework.sh
@@ -1,0 +1,189 @@
+#!/usr/bin/env sh
+
+NAME="aries_askar"
+VERSION="0.2.8-dev.1"
+BUNDLE_IDENTIFIER="org.hyperledger.$NAME"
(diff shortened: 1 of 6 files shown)


build-xcframework.sh
```diff
@@ -0,0 +1,189 @@
+#!/usr/bin/env sh
+
+NAME="aries_askar"
+VERSION="0.2.8-dev.1"
```
andrewwhitehead, line 4: Can we pull the version number from elsewhere to avoid having another place to update it?

blu3beri:
It now uses jq to fetch the version from the `Cargo.toml`.

TimoGlastra (approved):
Is jq installed by default?

blu3beri:
> Is jq installed by default?

Yeah that's the one thing I was not sure about. I will test it quickly on a random project.

blu3beri:
@TimoGlastra yes jq is by default installed on GH runners :).

PR #96, opened 2023-02-10
Add Swift wrapper

### LINKED ISSUE #99: cbindgen does not generate some types
cbindgen does not generate some types such as StoreHandle, SessionHandle, ScanHandle, which are types created using a macro.
It also omits `askar_string_free()` function.

conanoc:
This is a PR to add a Swift wrapper.
- Modified cbindgen.toml to generate types created with macro. (Fix #99)
- Others are Swift wrapper codes.
- Note: A Swift package requires the xcframework binary to be uploaded somewhere, so I uploaded it to aries-framework-swift repo. This Swift package will be hosted by this repo, but the xcframework binaries can be hosted by any repo.


build_xcframework.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
blu3beri, line 1: I think in `main` now there is a full xcframework build script that actually uses dynamic libraries. I had had quite some issues where iOS or Android could not find the correct symbols on static libraries and this would keep the build a bit smaller.

conanoc: I was unaware that the script had been added to the repository. I think I can remove mine and use that one. I'll submit a patch.

conanoc:
Can you review this PR @blu3beri @andrewwhitehead ?

include/libaries_askar.h
```diff
@@ -7,7 +7,7 @@
```
blu3beri, line 7: This is a generated file so if something changes here we should make sure that cbindgen generates the file with these changes as this will be overwritten when a new header is generated. Could you make it gets generated with your changes?

wrappers/swift/Askar/Package.swift
```diff
@@ -0,0 +1,29 @@
...
+    targets: [
+        .target(
+            name: "Askar",
+            dependencies: ["AskarFramework"]),
+        .testTarget(
+            name: "AskarTests",
+            dependencies: ["Askar"]),
+        .binaryTarget(
+            name: "AskarFramework",
+            url: "https://github.com/hyperledger/aries-framework-swift/releases/download/binary-release-askar-v0.2.7/AskarFramework.xcframework.zip",
```
blu3beri, line 25: Is this url definitive? There is already an xcframework on the releases of this library. Maybe in the future it can just use that one (right now it also includes Android builds though).

blu3beri (approved):
Some minor notes, but mainly the header generation code is important as the next time its generated all the changed will be reverted.

I am not a Swift developer so I did not review the swift code in here :).

build-xcframework.sh
```diff
@@ -125,7 +127,7 @@ cp ../../$HEADER_PATH/$HEADER_NAME Headers/$FRAMEWORK_LIBRARY_NAME.h
 mkdir Modules
 touch Modules/module.modulemap
 cat <<EOT >> Modules/module.modulemap
-framework module $FRAMEWORK_LIBRARY_NAME {
+framework module $FRAMEWORK_MODULE_NAME {
```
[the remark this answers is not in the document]
conanoc: It's the name of the module used in Objective-C or Swift. I want it to be imported with the name "AriesFramework" instead of "aries_askar". See https://github.com/conanoc/aries-askar/blob/swift_pr/wrappers/swift/Askar/Sources/Askar/Crypto.swift#L2
It's the convention in Swift, and this will not affect other languages.

wrappers/swift/Askar/Package.swift
conanoc: This link should be changed later. I'm thinking of adding a workflow job for this.

include/libaries_askar.h
conanoc: I'm not sure if it's possible. I've created an issue (#99) to track this.

build-xcframework.sh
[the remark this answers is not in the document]
conanoc: I was wrong. Swift does not seem to consult this module name when importing a framework. I'll rollback this.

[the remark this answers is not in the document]
conanoc: done

include/libaries_askar.h
conanoc: done

conanoc:
@andrewwhitehead I get the following error on pg_test. What should I do?
> error: package `csv v1.2.0` cannot be built because it requires rustc 1.60 or newer, while the currently active rustc version is 1.58.0

andrewwhitehead:
The build issue should be fixed by #102

conanoc:
@andrewwhitehead Could you merge this and #102 PR?

conanoc:
I extracted part of this PR to #110 and I'll close this.

PR #103, opened 2023-02-24
Set cross base images for manylinux support

andrewwhitehead:


andrewwhitehead:
@dbluhm I think this should produce more compatible wheels for Linux? I'm seeing `consistent with the following platform tag: "manylinux_2_17_x86_64` rather than `manylinux_2_31_x86_64`.

blu3beri (approved):
LGTM! Will these changes, where applicable, also be done for indy-vdr and Anoncreds?

andrewwhitehead:
> LGTM! Will these changes, where applicable, also be done for indy-vdr and Anoncreds?

Yes, I'd like to update those two and indy-shared-rs ideally.

PR #105, opened 2023-02-27
fix(js): empty key entry list metadata in react-native

genaris:
This prevents a crash  when fetching an existing key in iOS (and likely also in Android) and `askar_key_entry_list_get_metadata` returns NULL:

image

Not sure if the same should be applied to other parts of the code (it was already done in some other methods such as `entryListGetTags` and  `keyEntryListGetTags`).


wrappers/javascript/aries-askar-react-native/cpp/ariesAskar.cpp
```diff
@@ -1016,7 +1016,7 @@ jsi::Value keyEntryListGetMetadata(jsi::Runtime &rt, jsi::Object options) {
       askar_key_entry_list_get_metadata(keyEntryListHandle, index, &out);
   handleError(rt, code);
 
-  return jsi::String::createFromAscii(rt, out);
+  return jsi::String::createFromAscii(rt, out ? out : "{}");
```
blu3beri, line 1019:
```suggestion
  if out == nullptr {
    return jsi::Value::undefined();
  } else {
    return jsi::String::createFromAscii(rt, out);
  }
```

blu3beri (approved):
I am fine with either, but I think returning an undefined would be better instead of an empty string. I hope the syntax is correct :).

wrappers/javascript/aries-askar-react-native/cpp/ariesAskar.cpp
genaris: I changed the return type as suggested, and did the same with `entryListGetValue` and `entryListGetTags`. Seems to be working fine with the RN sample app from [test repo](https://github.com/TimoGlastra/shared-components-react-native-example).

blu3beri (approved):

PR #106, opened 2023-03-01
Update version to 0.2.8

andrewwhitehead:


swcurran (approved):

andrewwhitehead:
Maybe #105 should be merged first?

swcurran:
I was approving the changes — not necessarily that it be merged immediately :-).

@blu3beri @genaris @TimoGlastra @dbluhm — are there any other unmerged PRs (such as #105) that we should merge before this final 0.2.8 PR is merged?

genaris:
#105 and #107 are related to some issues found during integration with React Native environment. There is also at least another remaining fix for this platform (erro thrown when attempting to open a non-existant wallet). If @blu3beri has some time to quickly review them and they are fine, I think it would be nice to incorporate them to this release. 

That being said, as we are still testing it, it's likely that we'll have some other issues, so we might need to do further fixes anyway.

blu3beri:
I would like to get #104 in. If we cannot do this then it would be good for the JS wrapper if we do a 0.2.9 release rather quickly afterwards as our integration into Bifold and finishing up the 0.4.0 release depends on the migration script.

TimoGlastra:
Yeah if we can get the migration script in, it would then be possible to keep iterating on the wrapper dev versions while the 0.2.8 binary is already released.

PR #107, opened 2023-03-02
fix(js): free key and entry objects

genaris:
As there is not a destructor in JS, objects must be freed manually. This PR attempts to free any native object after it is used, making life easier to consumer applications. 

If this approach is correct, the only objects an app using this library should care about is `Key` (for which I added the missing call to `keyFree`).

@blu3beri let me know if concepts applied here are correct so I can replicate for anoncreds-rs.


andrewwhitehead:
It looks like EntryLists are freed in the Scan implementation which is good as well. It seems fairly easy for a user to accidentally leak sessions and run out of active connections though, it might be better to use a callback interface for working with open sessions and transactions, and close automatically after it runs.

blu3beri (approved):

Issue #77, opened 2023-01-20
[JS Wrapper] Custom error thrown when no entries found


genaris:
There are certain use cases where Askar JS is throwing a custom error (code 100) when I think it would be more appropriate to throw a specific error. These are the cases I've found during an initial integration with aries-framework-javascript:

- **Session.fetch** and no entry has been found (should be NotFound or simply return undefined, as current API suggests)
- **Session.fetchKey** and no entry has been found (should be NotFound as current API suggests)
- **Scan.fetchAll**/Session.fetchAllKeys and no entries matching the query have been found (should be NotFound or simply return an empty array)

TimoGlastra:
This is fixed by https://github.com/hyperledger/aries-askar/pull/111

PR #108, opened 2023-03-04
fix(js): ensure scan is freed after fetchAll

genaris:
When using `Scan.fetchAll` I found that the scan handle was not always properly freed. It turns out that this had to do with an error thrown when calling `scanNext` due to no records have been found (related issue https://github.com/hyperledger/aries-askar/issues/77), which prevented the code from executing the scanFree call.

So here we simply add a try-finally to make sure it is always being called.


TimoGlastra (approved):

TimoGlastra:
We have fixed the error being thrown when no record being found (in a local branch), but this fix is good either way

PR #110, opened 2023-03-06
Support for Swift wrapper

### LINKED ISSUE #99: cbindgen does not generate some types

conanoc:
This fixes #99, which is necessary for Swift.

The xcframework built for reactive-native can be shared with the Swift wrapper. But, it requires a different archive format. So I added a workflow step to archive it using `zip` with a different folder name `AskarFramework`.
The Swift wrapper in different repo could download the zipped xcframework for its dependencies.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -524,3 +524,15 @@ jobs:
         with:
           file: library-ios-android.tar.gz
           asset_name: "library-ios-android.tar.gz"
+
+      - name: Package for Swift
+        run: |
+          mv mobile/ios/aries_askar.xcframework AskarFramework.xcframework
+          zip -r AskarFramework.xcframework.zip AskarFramework.xcframework
+          shasum -a 256 AskarFramework.xcframework.zip
+
+      - name: Add xcframework to release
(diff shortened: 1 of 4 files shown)


andrewwhitehead:
Maybe we should just publish the xcframework as a zip and the Android libraries as a separate .tar.gz?

conanoc:
> Maybe we should just publish the xcframework as a zip and the Android libraries as a separate .tar.gz?

Sounds good. @blu3beri What do you think?

conanoc:
I think I need to test uniffi first because uniffi will generate a different header file. I'll close this now and re-open it later if uniffi fails.

TimoGlastra:
> Maybe we should just publish the xcframework as a zip and the Android libraries as a separate .tar.gz?

We need the ios and android build to be published in one tar/zip so they can be dynamically downloaded when the react native package is installed from npm. We can't download multiple binaries, or we'd have to implement that logic ourselves

Issue #109, opened 2023-03-05
SQLite databases configuration for mobile environments


genaris:
While testing Askar in mobile environments for integration with [AFJ](https://github.com/hyperledger/aries-framework-javascript), I had a problem in Android that resulted to be related to `max_connections` default value: Askar chooses this value depending on the number of CPUs, and as I was using an emulator with only 1 CPU available, there it was allowing at most a single connection to be opened. 

In AFJ, we open a session when we open the wallet and close it when agent shuts down. We use this session for all fetches and updates, including those about keys. This works fine for us until we do a query using `Scan`, because it is opening a new connection and of course will not be allowed under these circumstances, resulting in a pool timeout. That makes me think that probably we are not using Askar as it should: I mean, is it OK to do what we do (open a session and use it for all transactions except of scans), or should we always open and close connections on-demand (that's it, per single fetch/update)?

Another doubt I have is about configuration of **WAL mode** in SQLite, which is enabled by default. I think there is not any limitation on mobile environments on using this mode, but something I noticed during this early testing is that some transactions are lost when app is not closed gracefully (so the connection is not explicitly closed). This happened in cases where app crashed, but I'm a bit worried about what will happen when app goes to background and OS kills it (it's not always easy to get a callback to an app destructor when using environments like React Native). So the question is... would it be safer to simply disable WAL mode if we use Askar on mobile apps?

andrewwhitehead:
It sounds like you might be using a single transaction which isn't always committed at shutdown? This could lead to data loss. Otherwise, I don't think that WAL mode should lead to issues. There are some existing tests for contention between transactions ensuring there is no data loss or deadlocks occurring under normal usage, assuming transactions are always committed at the end.

I think in general it should be okay to use a single session for all those operations, although you would lose the ability to perform multiple operations transactionally because there are no nested transactions at the moment. The SQLite backend should be updated to have a minimum of 2 or 3 connections rather than being as low as 1 to help avoid deadlocks between connections.

genaris:
Thanks for the explanation (and the fix!) @andrewwhitehead !

Now that #114 is merged, it is working fine without need of specifying `max_connections`. :rocket:

PR #114, opened 2023-03-07
Restrict default max conns to a range for sqlite

andrewwhitehead:
Now using a default maximum number of connections between 2 and 8 even when the detected number of CPUs is higher or lower.

Related: #109


TimoGlastra (approved):

TimoGlastra:
Whoops didn't see CI failed

PR #123, opened 2023-03-16
fix(js): downcast with static cast

blu3beri:


TimoGlastra (approved):
tested in iOS & Android

PR #124, opened 2023-03-17
chore(js): update version

blu3beri:


TimoGlastra (approved):

TimoGlastra:
this is missing the binary and python version. We've kept them all in sync until now

blu3beri:
> this is missing the binary and python version. We've kept them all in sync until now

Yeah they out of sync, the last release was also out of sync. They will go out of sync anyways when we release them , so I did not see the harm in it.

TimoGlastra:
It's a bit weird I think, especially since we've done it for almost all releases and we also do it like this in the other shared component libraries

PR #127, opened 2023-04-03
fix(js)!: add argon2i(mod|int) support

blu3beri:

- StoreKeyMethod can now be properly created with `argon2i`


wrappers/javascript/aries-askar-shared/src/store/StoreKeyMethod.ts
```diff
@@ -0,0 +1,42 @@
...
+  Interactive = 'int',
+  Moderate = 'mod',
+}
+
+export class StoreKeyMethod {
+  private method: KdfMethod
+  private argon2Level?: Argon2Level
+
+  public constructor(method: KdfMethod, argon2Level?: Argon2Level) {
+    if (method == KdfMethod.Kdf && !argon2Level) {
```
TimoGlastra, line 19:
```suggestion
    if (method === KdfMethod.Kdf && !argon2Level) {
```

TimoGlastra: Should we add a check if kdf level is Raw and argon2level is defined? Also should we just add more enum values instead? What's the benefit of having kdf and then mod/int instaed of having kdf_mod and kdf_int?

TimoGlastra (dismissed):

wrappers/javascript/aries-askar-shared/src/store/StoreKeyMethod.ts
blu3beri: Mainly did it to be inline with how it is done in askar itself and this allows, although not quite good, more kdfs.

blu3beri: I will change it to `Argon2IMod` and `Argon2IInt` as growing with more kdfs would also not be a breaking change then.

wrappers/javascript/aries-askar-shared/src/store/StoreKeyMethod.ts
```diff
@@ -0,0 +1,18 @@
...
+  private method: KdfMethod
+
+  public constructor(method: KdfMethod) {
+    this.method = method
+  }
+
+  public toUri() {
+    return this.method.toString()
+  }
+}
```
TimoGlastra, line 18: Is this for future proofing? We could just use the enum value directly right?!

TimoGlastra (approved):

wrappers/javascript/aries-askar-shared/src/store/StoreKeyMethod.ts
blu3beri: Yes, it is for future proofing. Argon could have a salt or other kdf methods might need some custom logic.

PR #128, opened 2023-04-04
chore(js): update version

blu3beri:
- When this is merged I will do a "publish wrappers" GH action.


blu3beri:
This is needed for argon2i support for the JS wrapper.

PR #132, opened 2023-04-27
fix(nodejs): compatible with Windows builds

### LINKED ISSUE #131: Binary uploaded to github release for windows has wrong name
This means it can't download it when you install this package for Node.JS on windows.

This is also the case for indy-vdr and anoncreds-rs.

Options are to:
- rename the file that is uploaded to release
- modify the script already used by node-gyp to determine the name

blu3beri:
closes #131


TimoGlastra (approved):

TimoGlastra:
Should we rename the previous binaries? That way we don't have to do a re-release now

Issue #135, opened 2023-04-28
[React Native] Minimum iOS version requirement does not match with Podspec


AlexanderShenshin:
I got error related to iOS versions support when trying to validate app bundle for Testflight upload with minimum iOS version set to 12.0:

```
Invalid Bundle. The bundle AppName.app/Frameworks/aries_askar.framework does not support the minimum OS Version specified
in the Info.plist.
```

After some investigation I found that `minos` value in binary is set to 16.2:

```
aries_askar.framework % vtool -show aries_askar 
aries_askar:
Load command 10
      cmd LC_BUILD_VERSION
  cmdsize 32
 platform IOS
    minos 16.2 <---------------
      sdk 16.2
   ntools 1
     tool LD
  version 820.1
Load command 11
      cmd LC_SOURCE_VERSION
  cmdsize 16
  version 0.0
```

iOS version specified in `aries-askar.podspec`:
```
s.platforms    = { :ios => "12.0" }
```

After I switched app minimum iOS version to 16.2, app bundle validation completes successfully.

This problem is also relevant for `indy-vdr` and `anoncreds` RN packages.

AlexanderShenshin:
@TimoGlastra That's the issue that I mentioned on a AFJ call yesterday.

Issue #134, opened 2023-04-28
[React Native] Missing MinimumOSVersion value in iOS package


AlexanderShenshin:
Missing `MinimumOSVersion` value in `Info.plist` causing iOS app bundle validation to fail (Testflight upload):

```
Invalid MinimumOSVersion. Apps that only support 64-bit devices must specify a deployment target of 8.0 or later. MinimumOSVersion
in 'AppName.app/Frameworks/aries_askar.framework' is ".
```

Can be fixed locally by manually setting value in `Info.plist` file.

This problem is also relevant for `indy-vdr` and `anoncreds` RN packages.

Issue #133, opened 2023-04-28
[React Native] Invalid character in iOS bundle identifier


AlexanderShenshin:
There is an error on validation of iOS app bundle archived for Testflight upload:

```
This bundle is invalid. The bundle at path Payload/AppName.app/Frameworks/aries_askar.framework has an invalid
CFBundleldentifier 'org.hyperledger.aries_askar' There are invalid characters (characters that are not dots, hyphen and alphanumerics)
that have been replaced with their code point 'org.hyperledger.aries\u005faskar' CFBundleldentifier must be present, must contain
only alphanumerics, dots, hyphens and must not end with a dot.
```

The reason is `_` character in `org.hyperledger.aries_askar` identifier.

It's possible to work around this by updating local `Info.plist` file at `node_modules/@hyperledger/aries-askar-react-native/native/mobile/ios/aries_askar.xcframework/ios-arm64/aries_askar.framework`:

```
<key>CFBundleIdentifier</key>
<string>org.hyperledger.ariesaskar</string>
// "_" character was removed
```

Just to clarify, local build and runtime functionality are not affected by this - it's related to Apple validation rules for app distribution.

Such problem is also relevant for `indy-vdr` RN package.

swcurran:
@blu3beri — can you take this and issues #134 and #135?  They seem to be related.

PR #136, opened 2023-04-30
fix(ios): lower version compatibility and improved Info.plist

blu3beri:
@AlexanderShenshin I changed some of the values regarding the issues you created (#133 , #134 , #135). 

Hopefully this will suffice for all the issues.

> NOTE: if this works, I will do the same for indy-vdr and anoncreds.

### DIFF
--- a/wrappers/javascript/aries-askar-react-native/aries-askar.podspec
+++ b/wrappers/javascript/aries-askar-react-native/aries-askar.podspec
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
-  s.platforms    = { :ios => "12.0" }
+  s.platforms    = { :ios => "14.0" }
   s.source       = { :git => "https://github.com/hyperledger/aries-askar", :tag => "#{s.version}" }
 
   s.header_mappings_dir = "cpp"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -371,16 +371,9 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@master
         with:
-          # First version with support for cdylib: https://github.com/rust-lang/rust/pull/100636
-          toolchain: "1.65.0"
+          # Adds support for iOS version 13 and up
+          toolchain: "1.69.0"
           targets: ${{ matrix.target }}
-
-      # Not useful unless the toolchain version matches the default
-      # - name: Cache cargo resources
-      #   uses: Swatinem/rust-cache@v2
-      #   with:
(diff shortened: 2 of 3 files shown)


wrappers/javascript/aries-askar-react-native/aries-askar.podspec
```diff
@@ -10,7 +10,7 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
-  s.platforms    = { :ios => "12.0" }
+  s.platforms    = { :ios => "14.0" }
```
TimoGlastra, line 13: The comment in the other file mentions 13 and up. What will this support? Could we easily support lower versions? 14 is still quite new-ish (and higher than minimum bifold version)

blu3beri: I could not find any metrix that links rust to iOS version. I will try 1.60.x - 1,68.2 and see which has the lowest version compatibility.

blu3beri: Hmmm, I tried to look for a lower build version and all seem to target 16.2. I think I misread the 13 from somewhere...

They added dylib support in rust for iOS, back, at 1.65.0. I checked all versions above and they all have a `minos` version of 16.2 which is 67% of the iPhones (https://iosref.com/ios-usage). 

If we must have support for lower versions we need to move back to static libraries.

TimoGlastra: 16.2 is not acceptable

blu3beri: I also tested this without askar and just a `Hello World` example and they all also resolve to `16.2`.

blu3beri: > 16.2 is not acceptable

Okay. Back to static libraries it is.

genaris: This seems a bit weird to me, as one of my test handsets is an iPhone SE (1st generation) running iOS 15.7.3 and it somehow has been working so far with all askar package dev releases up to last week or so. What could be the reason for that?

AlexanderShenshin: Well, it looks like this issue does not affect any functionality in development. I'm also was able to run the app on iOS 14 in development, using Metro bundler.

And it's also possible  to create a iOS app bundle, but the problem is that it will fail bundle validation required to distribute the app (for example, upload the app to Apple Testflight). 

For details, please see https://github.com/hyperledger/aries-askar/issues/135.

PR #137, opened 2023-05-05
fix(js): scan fetch record limit

genaris:
There is an unnecessary chunk variable (inherited from previous Indy SDK implementation in AFJ) that is preventing `fetchAll` to retrieve all records when they are more than PAGE_SIZE. As a result, a given Scan cannot return more than 32 entries.

Simply checking for `listHandle` (`scanNext` return value) validity seems to be enough to determine if there are no more matching records.

### DIFF
--- a/wrappers/javascript/aries-askar-shared/src/store/Scan.ts
+++ b/wrappers/javascript/aries-askar-shared/src/store/Scan.ts
@@ -58,8 +58,6 @@ export class Scan {
     }
 
     try {
-      // Allow max of 256 per fetch operation
-      const chunk = this.limit ? Math.min(256, this.limit) : 256
       let recordCount = 0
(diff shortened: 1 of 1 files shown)


TimoGlastra (approved):
Isn't the 256 here to prevent you from retrieving too many records and messing stuff up? Or do you we want to make that a responsiblity of the caller?

genaris:
> Isn't the 256 here to prevent you from retrieving too many records and messing stuff up? Or do you we want to make that a responsiblity of the caller?

I think in this case is a bit different from Indy SDK because `scanNext` does not receive a chunk size: instead, it is retrieving at most 32 records per call (in Indy we used `fetchWalletSearchNextRecords`, which received a count parameter). So the loop will be executed until specified limit is reached or a `scanNext` call returns 0 records.

PR #150, opened 2023-06-02
Feat: kotlin wrapper

KolbyRKunz:
Created a Kotlin MPP wrapper that should have the same functionality as the node.js wrapper.


conanoc:
Wow. Glad to see this PR. Could you try this with #173 ? You will be able to run your test code, with slight modification, without the wrapper.

swcurran:
Closing this PR. We've made it so that wrappers are in separate repos (e.g. https://github.com/openwallet-foundation/askar-wrapper-javascript), so if you are still interested in this, it would be in a separate repo.

PR #151, opened 2023-06-19
Updates to the README and other Admin files in the repository

swcurran:
Signed-off-by: Stephen Curran <swcurran@gmail.com>


swcurran:
@andrewwhitehead -- note the reference in the readme that link to the indy-wallet documents. Is it safe enough to reference those (with the caveat that I put in) or are we better to leave them off until we get an Askar-specific version of the documents?

What's up with this documentation only PR failing tests?  :-)

andrewwhitehead:
I think it's fine to reference indy-wallet, although the implementation details are not all the same. It's based on the specifications (and reverse engineering) of the indy-wallet format, not the code.

swcurran:
That’s what I was trying to say — sounds like I missed the mark a bit.  I’ll look to see if I can make that clearer.

swcurran:
Updated that phrase.  Good enough?

andrewwhitehead (approved):

PR #159, opened 2023-07-20
Support copying individual profiles or a whole Store

### LINKED ISSUE #18: Askar different import/export capability
Using Aries Askar with Aries cloud agent, we would like to support the export of the wallet for possible backup/restore flows.

Is there any work planned to support:
- export of the whole database
- export of a single profile

andrewwhitehead:
Fixes #18

Also adds methods to get and set the default profile (opened when none is specified).

### DIFF
--- a/askar-storage/src/backend/mod.rs
+++ b/askar-storage/src/backend/mod.rs
@@ -128,14 +128,13 @@ pub trait BackendSession: Debug + Send {
     /// Insert scan results from another profile or store
     fn import_scan<'q>(
         &'q mut self,
-        kind: EntryKind,
         mut scan: Scan<'q, Entry>,
     ) -> BoxFuture<'_, Result<(), Error>> {
         Box::pin(async move {
             while let Some(rows) = scan.fetch_next().await? {
                 for entry in rows {
                     self.update(
-                        kind,
+                        entry.kind,
                         EntryOperation::Insert,
                         entry.category.as_str(),
                         entry.name.as_str(),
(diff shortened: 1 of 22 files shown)


TimoGlastra (approved):
These changes seem very useful! 

Need to make sure we also add these changes to the JS wrapper, but the changes surface of API seems quite small.

One question: Something in noticed with profile is that if I open a session for a profile that does not exist, it will not throw an error until a record is actually inserted of fetched from that session. This means I have to check whether a profile exists before opening, which is possible by calling `createProfile`. Or we have to fetch something to see if it works. I see some changes to the implementation of profiles, does this PR affect that behvaiour in any way?

What we do with normal wallets is that we try to open them, and if it fails we create them. This allows us to open it directly in all cases except the first time. While now we have to do a check every time that will fail everytime except the first time. It would be useful if opening a session would fail if the profile does not exist yet. 

If this PR does not affect this at all, I can also open a separate issue for it

andrewwhitehead:
@TimoGlastra Thanks for opening the issue. This PR doesn't change that behaviour (although it adds some profile-related methods) but it's a good one to fix for 0.4.

askar-storage/src/backend/mod.rs
```diff
@@ -140,3 +173,49 @@ pub trait BackendSession: Debug + Send {
...
+
+/// Export an entire Store to another location
+pub async fn copy_store<'m, B: Backend, M: ManageBackend<'m>>(
+    source: &B,
+    target: M,
+    key_method: StoreKeyMethod,
+    pass_key: PassKey<'m>,
+    recreate: bool,
+) -> Result<(), Error>
+where
```
berendsliedrecht, line 210:
```suggestion
```

berendsliedrecht (approved):

PR #161, opened 2023-08-02
fix(js): storeRekey in React Native

genaris:
StoreRekey for React Native was not calling the proper method.

### DIFF
--- a/wrappers/javascript/aries-askar-react-native/cpp/ariesAskar.cpp
+++ b/wrappers/javascript/aries-askar-react-native/cpp/ariesAskar.cpp
@@ -179,16 +179,17 @@ jsi::Value storeGetProfileName(jsi::Runtime &rt, jsi::Object options) {
 
 jsi::Value storeRekey(jsi::Runtime &rt, jsi::Object options) {
   auto storeHandle = jsiToValue<int64_t>(rt, options, "storeHandle");
-  auto keyMethod = jsiToValue<std::string>(rt, options, "keyMethod");
+  auto keyMethod = jsiToValue<std::string>(rt, options, "keyMethod", true);
   auto passKey = jsiToValue<std::string>(rt, options, "passKey");
 
   jsi::Function cb = options.getPropertyAsFunction(rt, "cb");
   State *state = new State(&cb);
   state->rt = &rt;
 
-  ErrorCode code = askar_store_get_profile_name(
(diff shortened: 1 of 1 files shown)


genaris:
@thiagoromanos

thiagoromanos:
> Do we also want to make a new release right away

If it's possible, it would be great, @TimoGlastra. Key rotation is a essential functionality for a feature my team is developing.

genaris:
> Do we also want to make a new release right away

I guess so, as different teams encountered issues related to this recently. It would not require new release at AFJ side, right? I mean consumers of the framework can use `@hyperledger/aries-askar-react-native 0.1.1` if they want? (currently `@aries-framework/askar` has `@hyperledger/aries-askar-shared: "^0.1.0"` as peerDependency)

andrewwhitehead:
It would be nice to merge #159 before a release. I also had a minor update to the Python wrapper planned, and the next version should be 0.4 due to a bunch of dependency updates and some breaking changes to the Rust API.

swcurran:
Added some reviewers on #159 so we can get that merged along with this one for a release.

andrewwhitehead:
We probably should do one or two 0.4 dev releases.

swcurran:
@andrewwhitehead — who is “we” on this.  I don’t want this falling between the cracks.  Are you the one to do the dev releases, and is there anything preventing them from happening?

andrewwhitehead:
I can or Berendt and Timo have created releases in the past. It just means updating the version numbers appropriately (can be slightly tricky as the format for dev versions varies between Rust/Python/JS) and creating a release on Github.

TimoGlastra:
As there is changes to the native bundle that don't have updates yet to the JS wrapper, I'm going to create a patch release for the JS wrapper for now. This will keep the existing native code, as it's a JS package only change

Then we can update the JS wrapper to support the new features form the recent PRs and get that also released

TimoGlastra:
Made a PR to update the version for JS: https://github.com/hyperledger/aries-askar/pull/165, so at least this PR can be released without any updates to the native bundle

TimoGlastra:
v0.1.1 has now been released. I think it should automatically install the new version if you remove your yarn.lock and regenerate it, but you can also force it by adding the `v0.1.1` dependency for both `/sahred` and `/react-native` to your project

PR #165, opened 2023-08-04
chore: update version for js wrapper

TimoGlastra:
Updates version for the JS wrapper to 0.1.1 so we can make another release before updating the JS wrapper with the changes for the 0.3 release.

Also adds new options publish-javascript-wrapper and publish-python-wrapper instead of publish-wrappers in the CI, so we can release the JS wrapper separately.

Also adds a version compatibility between Aries Askar version and the JS wrapper version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,8 +22,13 @@ on:
         required: true
         default: false
         type: boolean
-      publish-wrappers:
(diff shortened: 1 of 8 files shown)


genaris (approved):
I'm all in for this change! You have my grey tick mark 😄

PR #173, opened 2023-08-25
Introduce uniffi module and Swift wrapper

conanoc:
This is a replacement for the previous PR #110 using [uniffi](https://github.com/mozilla/uniffi-rs) to support Swift.
This uniffi module could be used for other languages such as Kotlin later.

As we have discussed in the Discord channel, holding many wrappers in one repo is not preferred, so I want to get some advice if there's a way to make this uniffi module a separate package that uses aries-askar as a dependency in another repo.
Any comment will be appreciated.


conanoc:
I've tried extracting the uniffi module to a separate package elsewhere and I could build it successfully by changing the visibility of some structs and members of aries-askar.
So, I think it will be better to move the uniffi module elsewhere and make a new PR for the visibility changes only. What do you think? @andrewwhitehead

conanoc:
As you responded to the discord message, I'll close this and create a new PR for visibility change only. Thanks.

PR #175, opened 2023-09-05
Adjust permissions to support uniffi module outside the repo

conanoc:
Please take a look @andrewwhitehead 
I could make a uniffi module as a separate crate with these changes. I added some dummy documents in error.rs and store.rs to make them public.


src/kms/entry.rs
```diff
@@ -46,7 +46,7 @@ pub struct KeyEntry {
     /// Thumbprints for the key
     pub(crate) thumbprints: Vec<String>,
     /// Thumbprints for the key
-    pub(crate) tags: Vec<EntryTag>,
+    pub tags: Vec<EntryTag>,
```
andrewwhitehead, line 49: Note the tags are kept sorted to facilitate equality checks, and an outside caller might not maintain that. Can the KeyEntry just expose a reference to the tags?

src/lib.rs
```diff
@@ -4,7 +4,7 @@
 #![deny(missing_docs, missing_debug_implementations, rust_2018_idioms)]
 
 #[macro_use]
-mod error;
+pub mod error;
```
andrewwhitehead, line 7: Is this for the macro?

src/lib.rs
```diff
@@ -22,9 +22,9 @@ pub use askar_storage as storage;
 pub use askar_storage::future;
 
 #[cfg(feature = "ffi")]
-mod ffi;
+pub mod ffi;
```
andrewwhitehead, line 25: I suggested it might be better to copy the tags module to the uniffi crate, to avoid exposing the ffi module here. It's some duplication, but if this ffi interface is deprecated then it wouldn't matter.

src/lib.rs
```diff
@@ -22,9 +22,9 @@ pub use askar_storage as storage;
...
 
 pub mod kms;
 
-mod store;
+pub mod store;
```
andrewwhitehead, line 29: What's in the store module that's not being exposed?

src/kms/entry.rs
conanoc: Done

src/lib.rs
conanoc: No. It is used [here](https://github.com/conanoc/aries-askar/blob/uniffi_pr/src/uffi/error.rs#L3) to implement `From<Error>`.

src/lib.rs
conanoc: Done

src/lib.rs
conanoc: This is the symbols I use from store.
```pub use aries_askar::store::{entry, PassKey, Session, Store, StoreKeyMethod};```

Comment (andrewwhitehead) [reply src/lib.rs:29]: Can't you just `use aries_askar::{entry, PassKey, Session, Store, StoreKeyMethod}` though?

File: src/lib.rs (hunk 2 above)
Comment (andrewwhitehead) [reply src/lib.rs:7]: It should be possible to just import Error from the root of the crate as `aries_askar::Error`

File: src/lib.rs (hunk 3 above)
Comment (andrewwhitehead) [reply src/lib.rs:25]: I don't think this needs to be public in that case

File: src/kms/entry.rs [hunk 5]
```diff
@@ -65,6 +65,11 @@ impl KeyEntry {
         self.name.as_str()
     }
 
+    /// Accessor for the key tags
+    pub fn tags_as_ref(&self) -> &Vec<EntryTag> {
```
andrewwhitehead, line 69: This could return a slice to avoid exposing implementation details: `self.tags.as_slice()`

src/lib.rs
conanoc: Yeah. It works.

conanoc:
After all, What I need is just adding an accessor for `KeyEntry.tags`. I didn't know how rust symbols are exported. Thanks for your guide.

andrewwhitehead (approved):

PR #176, opened 2023-09-05
Update dependencies

berendsliedrecht:
depends on #170 

Updated lerna to newer version to remove dependabot security issues.


andrewwhitehead (approved):

TimoGlastra (approved):

genaris:
> Looks like it's having trouble resolving the ref-napi module?

I think it needs to be updated with base branch to get the latest changes from #170.

TimoGlastra (approved):

wrappers/javascript/yarn.lock
```diff
@@ -1288,6 +1143,7 @@
     "@types/ref-struct-di" "*"
 
 "@types/2060.io__ref-napi@npm:@types/ref-napi", "@types/ref-napi@*":
```
genaris, line 1145: CI still failing 😭. It looks like it will also need to clean yarn lock file + `yarn install` for it to properly fetch the new types dependencies.

wrappers/javascript/yarn.lock
```diff
@@ -1456,7 +1290,15 @@
   dependencies:
     "@types/ref-napi" "*"
 
-"@types/ref-struct-di@*", "@types/ref-struct-di@^1.1.6":
+"@types/ref-napi@*":
+  name "@types/2060.io__ref-napi"
```
genaris, line 1294: I think this replacement came from the previous types config that might be cached somewhere (@types/2060.io__ref-napi does not actually exist). I could solve this locally by deleting it from yarn.lock (or removing the yarn lock file entirely) and running `yarn install` again. That way it will retrieve the actual `@types/ref-napi` package as it should be in the first place.

PR #177, opened 2023-09-07
Postgres provisioning fixes

### LINKED ISSUE #129: Problems installing ACA-Py using Askar on Postgres 15+ on OpenShift/Container Orchestration Platforms
See [this issue in the ACA-Py repository](https://github.com/hyperledger/aries-cloudagent-python/issues/2199) for details on the problem and suspected cause.  It appears what is needed is a tweak to how the Askar storage creation process is executed.

Using Askar on Postgres 15 seems to be fine.

Once completed, please add a comment to the ACA-Py issue so we can upgrade the Askar dependency as needed.

andrewwhitehead:
For support of Postgres 15 where the CREATE permission isn't automatically granted on the 'public' schema.

Adds a new optional 'schema' parameter for postgres connection URLs (ie. `postgres://user:pw@host/db?schema=public`). When not set, the schema now defaults to the username when creating new tables.

For new stores,
- If the database is newly created, then the database owner is set explicitly
- A new schema is created in the database, with the same name as the user. This is consistent with the default search path: `('$user', public)`
- All tables are created in the specified schema

For existing stores,
- If the schema isn't specified, then all schemas in the search path are checked for the config table (not just 'public')
- When the schema is set explicitly in the connection URL, the search path is restricted just to that schema

Fixes #129 ?

### DIFF
--- a/askar-storage/src/any.rs
+++ b/askar-storage/src/any.rs
@@ -263,7 +263,7 @@ impl<'a> ManageBackend<'a> for &'a str {
             let opts = self.into_options()?;
             debug!("Open store with options: {:?}", &opts);
 
-            match opts.schema.as_ref() {
+            match opts.scheme.as_ref() {
                 #[cfg(feature = "postgres")]
                 "postgres" => {
                     let opts = postgres::PostgresStoreOptions::new(opts)?;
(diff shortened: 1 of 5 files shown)


TimoGlastra (approved):

andrewwhitehead:
#178 should fix the test failure on Windows and it might be best to merge that PR first.

TimoGlastra (approved):

PR #178, opened 2023-09-08
Add method to ping store sessions; verify profile existence when opening a session

### LINKED ISSUE #163: Opening a session for a non existent profile does not throw an error
Something in noticed with profile is that if I open a session for a profile that does not exist, it will not throw an error until a record is actually inserted of fetched from that session. This means I have to check whether a profile exists before opening, which is possible by calling `createProfile`. Or we have to fetch something to see if it works. I see some changes to the implementation of profiles, does this PR affect that behvaiour in any way?

What we do with normal wallets is that we try to open them, and if it fails we create them. This allows us to open it directly in all cases except the first time. While now we have to do a check every time that will fail everytime except the first time. It would be useful if opening a session would fail if the profile does not exist yet. 

_Originally posted by @TimoGlastra in https://github.com/hyperledger/aries-askar/pull/159#pullrequestreview-1560604404_

andrewwhitehead:
Fixes #163 

For Postgres, this also ensures that a profile cannot be removed while there is an active transaction against it.

Includes a fix for the persistent test failure copying a Store instance on Windows.

### DIFF
--- a/askar-storage/src/any.rs
+++ b/askar-storage/src/any.rs
@@ -239,6 +239,11 @@ impl BackendSession for AnyBackendSession {
             .update(kind, operation, category, name, value, tags, expiry_ms)
     }
 
+    /// Test the connection to the store
+    fn ping(&mut self) -> BoxFuture<'_, Result<(), Error>> {
+        self.0.ping()
+    }
+
     /// Close the current store session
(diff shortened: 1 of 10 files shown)


andrewwhitehead:
#178 should fix the test failure on Windows and it might be best to merge that PR first.

TimoGlastra (approved):

PR #180, opened 2023-09-15
Update to dalek 2.x crates

andrewwhitehead:


askar-crypto/Cargo.toml
```diff
@@ -48,32 +50,62 @@ harness = false
...
 blake2 = { version = "0.10", default-features = false }
 block-modes = { version = "0.9", default-features = false, optional = true }
-bls12_381 = { version = "0.8", default-features = false, features = ["groups", "zeroize"], optional = true }
+bls12_381 = { version = "0.8", default-features = false, features = [
+    "groups",
+    "zeroize",
+], optional = true }
 cbc = { version = "0.1", default-features = false, optional = true }
-chacha20 = { version = "0.9" }  # should match dependency of chacha20poly1305
+chacha20 = { version = "0.9" } # should match dependency of chacha20poly1305
```
berendsliedrecht, line 69: not a change you introduced, but `chacha20` does not match `chacha20poly1305`. Maybe we should remove the comment? The latest `chacha20` is 0.9.1 and `chacha20poly1305` is on `0.10.1`.

askar-crypto/src/alg/ed25519.rs
```diff
@@ -97,25 +93,14 @@ impl Ed25519KeyPair {
     /// Verify a signature against the public key
     pub fn verify_signature(&self, message: &[u8], signature: &[u8]) -> bool {
         if let Ok(sig) = Signature::try_from(signature) {
-            self.public.verify_strict(message, &sig).is_ok()
+            let vk = VerifyingKey::from_bytes(&self.public).unwrap();
```
berendsliedrecht, line 96: Is this infallable or should we return a `Result<bool>` for this?

askar-crypto/src/alg/ed25519.rs
```diff
@@ -135,10 +120,7 @@ impl Debug for Ed25519KeyPair {
 impl KeyGen for Ed25519KeyPair {
     fn generate(rng: impl KeyMaterial) -> Result<Self, Error> {
         let sk = ArrayKey::<U32>::generate(rng);
-        // NB: from_bytes is infallible if the slice is the right length
-        Ok(Self::from_secret_key(
-            SecretKey::from_bytes(sk.as_ref()).unwrap(),
-        ))
+        Ok(Self::from_secret_key((&sk).try_into().unwrap()))
```
berendsliedrecht, line 123: Is this still infallible? If so, might be good to leave the comment here.

berendsliedrecht (approved):
Couple minor comments about unwrapping.

askar-crypto/src/alg/ed25519.rs
andrewwhitehead: Same here, it would only fail if `ArrayKey<U32>` can't be turned into `&[u8; 32]`

askar-crypto/Cargo.toml
andrewwhitehead: This is correct, although the comment could maybe be clearer. Because we use both both of these crates (and there isn't a re-export available), it's important to make sure that the version of `chacha20` matches the version depended upon by `chacha20poly1305`.

Issue #195, opened 2023-11-02
Inefficient SQL fetch query


ff137:
Good day,

The following queries in [aries-askar/askar-storage/src/backend/postgres/mod.rs](https://github.com/hyperledger/aries-askar/blob/e53436215cddc5bb7d758b604e25d547c219b048/askar-storage/src/backend/postgres/mod.rs#L52C1-L65C75) appear to be source of performance degradation when a lot of wallet entries exist:
```rs
const FETCH_QUERY: &str = "SELECT id, value,
    (SELECT ARRAY_TO_STRING(ARRAY_AGG(it.plaintext || ':'
        || ENCODE(it.name, 'hex') || ':' || ENCODE(it.value, 'hex')), ',')
        FROM items_tags it WHERE it.item_id = i.id) tags
    FROM items i
    WHERE profile_id = $1 AND kind = $2 AND category = $3 AND name = $4
    AND (expiry IS NULL OR expiry > CURRENT_TIMESTAMP)";
const FETCH_QUERY_UPDATE: &str = "SELECT id, value,
    (SELECT ARRAY_TO_STRING(ARRAY_AGG(it.plaintext || ':'
        || ENCODE(it.name, 'hex') || ':' || ENCODE(it.value, 'hex')), ',')
        FROM items_tags it WHERE it.item_id = i.id) tags
    FROM items i
    WHERE profile_id = $1 AND kind = $2 AND category = $3 AND name = $4
    AND (expiry IS NULL OR expiry > CURRENT_TIMESTAMP) FOR NO KEY UPDATE";
```

The reason it is inefficient is coming from the subquery in the SELECT clause: `SELECT ARRAY_TO_STRING(ARRAY_AGG(...`

This subquery is executed for each row returned by the outer query. If the outer query returns a large number of rows, this can result in a significant performance hit.

Also, the subquery involves aggregating data and performing string manipulations, which are computationally expensive operations.

Now, I don't have an immediate solution to propose, but it will be something along the lines of replacing the subquery with a JOIN, and avoiding string manipulation/aggregation if possible.

Ideally the person who is most familiar with the SQL statements should be tasked with improving it, but in case of their absence, GPT-4 is your friend :-). Hopefully someone can propose a more efficient solution! Otherwise I'll try rehash my SQL skills when I have time available.

___
[...]

PR #196, opened 2023-11-06
Refactor Fetch SQL queries in Postgres backend

rblaine95:
An attempt at some optimizations on the `FETCH_QUERY` and `FETCH_QUERY_UPDATE` Postgres queries.

Opening as a draft PR while I work on figuring out a way to benchmark the updated queries.

According to GPT:
> The refactored versions with the LEFT JOIN and GROUP BY clauses, as well as the use of CTEs are likely to be more efficient in terms of performance, especially when dealing with large datasets. It allows for a more optimized approach to data retrieval, particularly when combining information from multiple tables.

Please be as pedantic as possible (`-Wclippy::pedantic -Wclippy::nursery`).

Related to #195

### DIFF
--- a/askar-storage/src/backend/postgres/mod.rs
+++ b/askar-storage/src/backend/postgres/mod.rs
@@ -49,20 +49,58 @@ const COUNT_QUERY: &str = "SELECT COUNT(*) FROM items i
     AND (expiry IS NULL OR expiry > CURRENT_TIMESTAMP)";
 const DELETE_QUERY: &str = "DELETE FROM items
     WHERE profile_id = $1 AND kind = $2 AND category = $3 AND name = $4";
-const FETCH_QUERY: &str = "SELECT id, value,
-    (SELECT ARRAY_TO_STRING(ARRAY_AGG(it.plaintext || ':'
-        || ENCODE(it.name, 'hex') || ':' || ENCODE(it.value, 'hex')), ',')
-        FROM items_tags it WHERE it.item_id = i.id) tags
-    FROM items i
-    WHERE profile_id = $1 AND kind = $2 AND category = $3 AND name = $4
-    AND (expiry IS NULL OR expiry > CURRENT_TIMESTAMP)";
-const FETCH_QUERY_UPDATE: &str = "SELECT id, value,
-    (SELECT ARRAY_TO_STRING(ARRAY_AGG(it.plaintext || ':'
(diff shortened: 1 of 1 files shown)


andrewwhitehead:
It appears that the latest ring dependency breaks support for the manylinux aarch64 environment we are using (https://github.com/briansmith/ring/issues/1728). We could update to `manylinux_2_28` or try one of the other fixes, I'll have to test that in another PR.

rblaine95:
Small update - I've made some progress on building a benchmark using [Criterion](https://crates.io/crates/criterion) to test if there's any benefit in the refactored SQL queries.

Preliminary results are looking good.
On my local (M2 Macbook Pro (10CPU, 16GiB)), I pre-populated the DB with 5000 profiles, then benchmarked the old SQL vs the new SQL creating 1000 profiles on top of the existing 5000.

Red = Old SQL
Blue = Refactored SQL
image

I'll work on a larger benchmark - pre-populate with 50000 profiles and then compare the SQL queries.
The expectation is that the refactored version _should_ be faster compared to the old version as the number of profiles increases.

askar-storage/src/backend/postgres/mod.rs
```diff
@@ -49,20 +49,58 @@ const COUNT_QUERY: &str = "SELECT COUNT(*) FROM items i
...
+            AND (i.expiry IS NULL OR i.expiry > CURRENT_TIMESTAMP)
+        GROUP BY
+            i.id,
+            i.value
+    )
+    SELECT
+        *
+    FROM
+        fetched_data
+    FOR UPDATE";
```
andrewwhitehead, line 103: Any reason this isn't using `FOR NO KEY UPDATE` now?

rblaine95: Great catch, thank you!

I didn't realize I had missed that.

rblaine95:
@swcurran, @andrewwhitehead, is there a channel on the Hyperledger Discord we could chat in?
I realized I may have been a little overly excited with my benchmark results earlier and that the tests/benchmarks aren't actually running in Multitenant mode. I'd like to figure out a way to pre-populate the DB with a number of profiles so that the benchmarking results are a little more appropriate.

swcurran:
There is an aries-askar channel on the Hyperledger Discord Server (invite: https://discord.gg/hyperledger) that I just created that we can use.  Should have been created long ago :-).

PR #198, opened 2023-11-09
Add workaround for building ring on aarch64

andrewwhitehead:


swcurran:
So close…can you make it work for Python too?

swcurran:
w00t!!!  Can we merge this? @berendsliedrecht @genaris

genaris:
Nice!!!

.github/workflows/build.yml
```diff
@@ -427,6 +432,11 @@ jobs:
       - name: Build
         run: |
           cargo install --bins --git https://github.com/rust-embedded/cross --tag v${{ env.CROSS_VERSION }} cross
+          # Required for compatibility with manylinux2014:
+          # https://github.com/briansmith/ring/issues/1728
+          if [ "${{ matrix.target }}" = "aarch64-linux-android" ]; then
+            export CFLAGS="-D__ARM_ARCH=8"
```
berendsliedrecht, line 438: Some android devices might still be on 7 IIRC. I strongly doubt this will cause any issues, but just leaving this here if we find issues in the feature regarding Android on ARMv7.

berendsliedrecht (approved):

PR #211, opened 2023-12-21
fix(js): promise handle error and rn updates

TimoGlastra:
Some additional updates after #209.

It addresses two things:
- also handle error code mismatch in the promise callback
- update the RN implementation to handle the errors in the same way as we updated to in the node js wrapper. @berendsliedrecht you mentioned it was already implemented correctly in RN, were you talking about the unnessary error calling? Because the code matching logic was not present in the RN wrapper.

I've changed the mismatch behaviour handling to now throw an error with the original errorCode, and a custom message "Error has already been overwritten, unable to provide error message". That way it won't break the calling implementation, as long as they use errorCodes to  handle errors. I've looked and it seems we're not using error message from Askar anymore to do error handling, so this should fix the problem, except for not always having the correct error message in the logs.

### DIFF
--- a/wrappers/javascript/aries-askar-nodejs/package.json
+++ b/wrappers/javascript/aries-askar-nodejs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hyperledger/aries-askar-nodejs",
-  "version": "0.2.0-dev.4",
+  "version": "0.2.0-dev.5",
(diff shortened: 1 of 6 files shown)


berendsliedrecht (approved):
I was talking about the "only call get error when an error actually occurred", yes. Not the error code mismatch

PR #239, opened 2024-04-16
Build Javascript on multiple platforms

brianorwhatever:


andrewwhitehead:
Windows support is skipped for now due to a MODULE_NOT_FOUND error involving node-pre-gyp (maybe it needs to be installed).

andrewwhitehead (approved):

PR #248, opened 2024-05-09
chore: update minSdkVersion to get the project version or fallback to 21

ja-bravo:
This PR updates the react-native wrapper to get the `minSdkVersion` from the project config or fallback to the current hardcoded version.

I'm not entirely sure if anything else is needed but with this change I could get an app running on React Native 0.74. This change would also be needed in othe repos (`@hyperledger/anoncreds-react-native` and `@hyperledger/indy-vdr-react-native` at least) but I'd like to get approval here before spamming the other repos 😄


TimoGlastra (approved):
This looks good to me. 

@berendsliedrecht could you also take a look?

berendsliedrecht (approved):

Issue #260, opened 2024-05-30
SQLite expiry compare is not correct.


dkulic:
In SQLite comparison of expiry date is done using query
```sql
(expiry IS NULL OR expiry > DATETIME('now'))
```
but SQLite considers text strings for date and time comparisons by character order, not by their actual date/time value., so the check is not working correctly and Askar returns the items stored even after expired.

The fix is easy, `expiry` should be first converted to DATETIME, in above query something like:
```sql
(expiry IS NULL OR DATETIME(expiry) > DATETIME('now'))
```

PR #261, opened 2024-05-30
fix: Incorrect SQLite expire check

dkulic:
In SQLite comparison of expiry date is done using query
```sql
(expiry IS NULL OR expiry > DATETIME('now'))
```
but SQLite considers text strings for date and time comparisons by character order, not by their actual date/time value., so the check is not working correctly and askar returns the items stored even after expired.

The fix is easy, `expiry` should be first converted to DATETIME, in above query something like:
```sql
(expiry IS NULL OR DATETIME(expiry) > DATETIME('now'))
```

Fixes issue #260


andrewwhitehead (approved):
Thanks!

PR #262, opened 2024-06-03
Add python requirements file and update dependabot workflow

ff137:
Currently setup.py just indicates one dependency: `cached_property`.

It makes sense if the dependencies were parsed from a requirements.txt file (less effort than migrating to pyproject.toml format). Allows for easier management if any extras need to be added, and also means we can add pip dependency management to the dependabot workflow (namely to catch if there's an update to `cached_property`.

A requirements.dev.txt file is also added to track dev / test dependencies. I would add indy as a test requirement, since it's used in `indy_test.py`, but I can't find out which `indy` packages it's referring to ... it doesn't work with `pip install indy` or `python3-indy`. So for now I've just specified `pytest` and `pytest-asyncio`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,6 +25,18 @@ updates:
       - dependency-name: "*"
         update-types: ["version-update:semver-major", "version-update:semver-patch"]
 
+  # Maintain dependencies for pip
+  - package-ecosystem: "pip"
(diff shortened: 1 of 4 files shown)


WadeBarnes (approved):
LGTM.  I'll defer to @andrewwhitehead about the `indy` dependency.

PR #264, opened 2024-06-04
Replace serde_cbor with ciborium

marlonbaeten:
The crate `serde_cbor` is unmaintained and has not received any updates in the last 3 years. The author of serde_cbor proposes [ciborium](https://crates.io/crates/ciborium) as an alternative.
This PR replaces the dependency on `serde_cbor` with `ciborium`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -159,6 +159,7 @@ dependencies = [
  "askar-crypto",
  "askar-storage",
  "async-lock",
+ "ciborium",
  "criterion",
  "env_logger",
  "ffi-support",
@@ -167,7 +168,6 @@ dependencies = [
  "once_cell",
  "rand",
  "serde",
- "serde_cbor",
  "serde_json",
  "zeroize",
 ]
@@ -189,6 +189,7 @@ dependencies = [
  "cbc",
  "chacha20",
  "chacha20poly1305",
+ "ciborium",
  "cipher",
  "criterion",
  "crypto_box",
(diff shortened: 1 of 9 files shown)


berendsliedrecht (dismissed):

ryjones:
@marlonbaeten @berendsliedrecht [take a look](https://github.com/hyperledger/aries-builder-images/pull/8) and see if this would work

andrewwhitehead:
It does look to me like the minimum API version only increased to KitKat which is already under 1% of devices, so I'm not sure how necessary this restriction is. Maybe it's for compatibility with other libraries? @berendsliedrecht

ryjones:
I suggest merging this and bumping the images repo to whatever is new enough.

TimoGlastra:
There were some issues we encountered without the custom images in quite recent Android APIs, not sure if that will impact it, but would be good to first test it thoroughly.

I think it was mainly the NDK version.

Berend would know more

berendsliedrecht:
> There were some issues we encountered without the custom images in quite recent Android APIs, not sure if that will impact it, but would be good to first test it thoroughly.
> 
> I think it was mainly the NDK version.
> 
> Berend would know more

IIRC it was mainly due to rust 1.67+ not working for the Android builds. But I am not too sure anymore if that was because of `cross` or a subdependency. 

I see that the CI fails on `error: package `half v2.4.1` cannot be built because it requires rustc 1.70 or newer, while the currently active rustc version is 1.67.1`. Which means we either have to run `cargo update -p half@2.4.1 --precise 2.2.1` and see if that fixes it or see if we can update building android images with 1.70 and fix the occurring errors for that.

andrewwhitehead:
Merged as part of #326, thanks!

PR #268, opened 2024-06-11
Removed Encryption from SQLite db 

vinaysingh8866:
Removed Encryption from SQLite db, 
Features working

- [x] Insert

- [x] Count

- [x] Replace

- [x]  Remove

- [ ]  Scan ( Partially working doesn't work with tag filters )

- [ ] Fetchall ( Partially working doesn't work with tag filters )

- [x] Key Insert

- [x] Fetch Key

- [ ] Fetch All Keys ( Partially working doesn't work with tag filters )


swcurran:
@vinaysingh8866 — what’s the goal of this PR?  Sounds like something we don’t want — removing encryption.

vinaysingh8866:
@swcurran oppsie accidentally created PR on this repo

PR #291, opened 2024-07-25
✨ add `order_by` and `descending` options to scan and fetch_all queries

### LINKED ISSUE #290: Support custom ordering in paginated queries
Currently, paginated querying of storage records does not take a custom ordering option, nor does it specify a default ordering in the SQL query being made. This means that DB shuffling or new records being added, can lead to duplicate or missing results across pages being returned. To resolve this, support for custom ordering should be added to the scan method.

ff137:
Closes #290

Adds ordering functionality to `scan` and `fetch_all` queries, to allow for custom ordering of fetched storage records.


ff137:
Ready for review!

I wanted to test compatibility with ACA-Py, but building and installing the python wrapper from my local branch, and running pytest in ACA-Py, gives me: 
```
INTERNALERROR>     get_library()
INTERNALERROR>   File "/root/anaconda3/envs/acapy/lib/python3.12/site-packages/aries_askar/bindings/__init__.py", line 44, in get_library
INTERNALERROR>     LIB.loaded
INTERNALERROR>   File "/root/anaconda3/envs/acapy/lib/python3.12/site-packages/aries_askar/bindings/lib.py", line 352, in loaded
INTERNALERROR>     self._lib = LibLoad(self.__class__.LIB_NAME)
INTERNALERROR>                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
INTERNALERROR>   File "/root/anaconda3/envs/acapy/lib/python3.12/site-packages/aries_askar/bindings/lib.py", line 106, in __init__
INTERNALERROR>     self._load_library()
INTERNALERROR>   File "/root/anaconda3/envs/acapy/lib/python3.12/site-packages/aries_askar/bindings/lib.py", line 129, in _load_library
INTERNALERROR>     raise AskarError(
INTERNALERROR> aries_askar.error.AskarError: Library not found in path: aries_askar
```
This resolves when I revert to installing official pypi package with pip install. So some extra step is needed when calling `pip install .` in the `aries-askar/wrappers/python` dir.

I tried adding `aries-askar/target/release` to `LD_LIBRARY_PATH`, but no dice.

If anyone can assist with testing, will be much appreciated!

ff137:
A logical extension would be to add the custom ordering option to "fetch all" queries as well (in a future PR)

ff137:
Good news! Managed to successfully test the changes locally 🚀 
after copying `target/release/libaries_askar.so` to `wrappers/python/aries_askar`. 🚚 

Happy to confirm that all the ACA-Py tests pass using this branch's changes. ✅ 
Will begin implementing the order_by functionality for ACA-Py soon™️

askar-storage/src/backend/db_utils.rs
```diff
@@ -453,6 +453,17 @@ pub trait QueryPrepare {
         }
         query
     }
+
+    fn order_by_query<'q>(mut query: String, order_by: Option<String>, descending: bool) -> String {
+        if let Some(order_by) = order_by {
+            query.push_str(" ORDER BY ");
+            query.push_str(&order_by);
```
andrewwhitehead, line 460: This order_by parameter needs to be sanitized, we don't want the caller to be able to inject raw SQL. Perhaps there should be an enum of allowed values? And do we want a default order (the ID I suppose)? In general the parameter might be a bit misleading as you can't sort by encrypted categories or values.

ff137: Great points.

> This order_by parameter needs to be sanitized, we don't want the caller to be able to inject raw SQL. Perhaps there should be an enum of allowed values?

I think an enum would be appropriate if the set of allowed values is known and the same across all different record types. I don't know yet what that possible set of values would be, but I can review the DB.

> And do we want a default order (the ID I suppose)? 

Default ordering by ID is probably a good idea

> In general the parameter might be a bit misleading as you can't sort by encrypted categories or values.

This is correct -- it is a bit misleading if one can't sort by encrypted record values. Part of the motivation for this implementation was to at least allow users to query by "last updated" or "created" timestamps. I hope those will be possible. But if they're encrypted then naturally it'll just have to be explained as a note in documentation, why it's not possible to order by other fields.

I'll take some time to think about best steps forward.

ff137: @andrewwhitehead I've implemented an Enum for OrderBy, and I noticed that it really only seems to make sense to ever order by ID. Please let me know if there are other fields that are generally safe to order by, which I can include in the enum. But for the time being, I think just ordering by ID is sufficient, to ensure consistency across multiple calls.

I've opted to preserved "backward compatibility" for now: if no OrderBy option is provided, there is no ordering, and the `descending` flag will also be ignored. I figured this is safer since I'm not sure if ID is a global field, in absolutely all records, so didn't want to force it to be always be default. If it is global, perhaps it's wise to make it default, even when no order by option is provided.

ff137: I also ensured that limit, offset, order_by, and descending, will only ever be added to SELECT queries. So that the `extend_query` changes don't impact DELETE or INSERT query types, where pagination doesn't make sense.

ff137:
I think I'm gonna try add the ordering options for fetching all records -- not just for scanning. 
Makes sense imo to include it all in one PR

ff137:
I've expanded ordering options to the `fetch_all` method as well.

I can confirm python wrapper works with changes (ACA-Py tests pass with this build). Not sure how to test javascript wrapper

ff137:
@andrewwhitehead Please let me know if you've had moment to review, if there are any additional steps necessary here. Maybe changing default behavior, or implementing some specific tests.

ACA-Py tests are passing with these changes here: https://github.com/hyperledger/aries-cloudagent-python/pull/3173

And our end-to-end tests, asserting unique results across pages, are passing now (as described here:
https://github.com/hyperledger/aries-cloudagent-python/pull/3173#issuecomment-2297491140)

andrewwhitehead (approved):
Thanks, I think it looks good!

PR #301, opened 2024-08-21
:construction: WIP: Replace default python `json` library with `orjson`

ff137:
json operations are notoriously expensive, especially with the default python library. `orjson` boasts 10x fastert de/serialisation speeds. So it'd be great if the python wrapper offered `orjson` an optional alternative.


andrewwhitehead:
Superseded by #327, although updating to use poetry and adding `orjson` as an optional dependency would still be nice.

PR #302, opened 2024-08-21
:sparkles: Python wrapper: use poetry

ff137:
Replaces the existing setup.py/setup.cfg/requirements.txt with pyproject.toml, and uses poetry in CICD build


ff137:
I realise I don't know enough about how the multi-architecture wheels are built, to replace things properly with poetry

andrewwhitehead:
I looked into this a bit but there seems to be no good way to pass through the `python-tag` and `plat-name` overrides for `bdist_wheel` when using `pyproject.toml`. Poetry doesn't seem to support them at all (there are open issues) and `pip` is looking at removing the `--build-option` flag which could allow it (if using `setuptools` as the build system).

PR #330, opened 2024-11-29
Implement JWK import for symmetric keys

### LINKED ISSUE #329: Unable to load symmetric key from storage in JS wrapper
Not sure if this is a problem in how we're fetching the key in the JS wrapper, or whether this hasn't been implemented yet in Askar.

Basically what we do is:
- create symmetric key (askar_session_insert_key)
- fetch the key (askar_session_fetch_key)
- load the key data into js (askar_key_entry_list_load_local)

The last method will return an error `JWK for key import`. The error seems to make sense as JWK representation for symmetric keys hasn't been implemented, (see https://github.com/openwallet-foundation/askar/blob/e902e7f0beb6a582d5c9ca838fee3d20d21fbcd3/askar-crypto/src/alg/any.rs#L631-L632) 

However, I would still suspect to be able to load the key, be it just for the secret bytes.

Is this an error in how we're loading the key (askar_key_entry_list_load_local), or an error in Askar?

andrewwhitehead:
Fixes #329

### DIFF
--- a/askar-crypto/src/alg/aes/mod.rs
+++ b/askar-crypto/src/alg/aes/mod.rs
@@ -127,7 +127,7 @@ impl<T: AesType> FromJwk for AesKey<T> {
         if jwk.kty != JWK_KEY_TYPE {
             return Err(err_msg!(InvalidKeyData, "Unsupported key type"));
         }
-        if jwk.alg != T::JWK_ALG {
+        if jwk.alg.is_some() && jwk.alg != T::JWK_ALG {
             return Err(err_msg!(InvalidKeyData, "Unsupported key algorithm"));
         }
         Ok(Self(ArrayKey::try_new_with(|buf| {
--- a/askar-crypto/src/alg/aes/mod.rs
+++ b/askar-crypto/src/alg/aes/mod.rs
@@ -13,7 +13,7 @@ use crate::{
     encrypt::{KeyAeadInPlace, KeyAeadMeta, KeyAeadParams},
     error::Error,
     generic_array::{typenum::Unsigned, GenericArray},
-    jwk::{JwkEncoder, ToJwk},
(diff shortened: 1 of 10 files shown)


swcurran:
@TimoGlastra — note that we have removed the JavaScript/RN wrapper from this repo. I’m organizing resources to get the `askar-wrapper-javascript` repo updated — still working on that.  Once this is merged, we can get a release of Askar (and the Python wrapper) completed, but we’ll need to do the work in the new repo to get the JavaScript artifacts generated.

TimoGlastra (approved):
Thanks for the quick fix! 

I see you added alg identifiers for XC20P. Is this based on some spec, i couldn't find anything for it?

andrewwhitehead:
@TimoGlastra No, I don't believe there's a spec defining those `alg` values, there also isn't an accepted one for BLS keys. The `C20P`/`XC20P` values do match the `enc` values proposed here: https://datatracker.ietf.org/doc/draft-amringer-jose-chacha/

TimoGlastra:
I think for BLS there is a newer one, which now represents them as `EC` keys instead of `OKP`. Might also be worth looking into: https://datatracker.ietf.org/doc/html/draft-ietf-cose-bls-key-representations

andrewwhitehead:
For BLS12-381 Askar currently accepts `OKP` or `EC` for the `kty`, although the `crv` values don't match that draft (difference of an underscore). And that draft is currently expired too...

TimoGlastra:
> For BLS12-381 Askar currently accepts `OKP` or `EC` for the `kty`, although the `crv` values don't match that draft (difference of an underscore). And that draft is currently expired too...

Ah interesting. Any way to trigger getting the EC variant? Because the OKP one only has X

PR #334, opened 2024-12-10
Update publish action

ryjones:
This needs the logic removed from build.yml (the actual publishing part, not the release creating part)

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -1,0 +1,29 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
+      url: https://pypi.org/p/aries-askar
+    permissions:
+      id-token: write  # IMPORTANT: this permission is mandatory for trusted publishing
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up Python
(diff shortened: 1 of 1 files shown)


andrewwhitehead:
The logic in `build.yml` publishes packages for multiple platforms. We also don't support poetry (yet) although there's a PR starting that work.

swcurran:
@andrewwhitehead — could you explain what your comment above means — what are the ramifications of what you are saying.  I think I’m reading it that we need to remove the multiple platforms and the poetry update needs to be completed?

jamshale:
I don't think this publishing method requires poetry. It's just how it's configured in pypi and with github actions. In theory the pypi publishing can just be removed from the build.yml pipeline and it can publish after, leaving the other platforms unchanged.

I haven't had the chance to study this yet though, so I could be off.

ryjones:
> I don't think this publishing method requires poetry. It's just how it's configured in pypi and with github actions. In theory the pypi publishing can just be removed from the build.yml pipeline and it can publish after, leaving the other platforms unchanged.
> 
> I haven't had the chance to study this yet though, so I could be off.

That was my thought - the completion of the other job could trigger the pypi job

jamshale:
I'm working on another PR for this...

swcurran:
@jamshale did a separate PR that handled the changes needed, and @andrewwhitehead updated the PyPi settings, so we should be good to go -- without this.  We'll see how the release goes...

PR #335, opened 2024-12-11
Update python publish workflow

jamshale:
This should work for the python publish. 

Currently on my forked branch I get:

> ERROR    HTTPError: 403 Forbidden from https://upload.pypi.org/legacy/          
>          Invalid API Token: OIDC scoped token is not valid for project          
>          'aries-askar', project-scoped token is not valid for project:          
>          'aries-askar'        

but this is because the github action on my fork isn't in the repo matching aries-askar. It's a lot of work to rename everything to test with a personal pypi project, so I haven't done that.

I'm trying to figure out if can share this `environment` config amongst all the matrixes but I can't find much.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -211,6 +211,8 @@ jobs:
   build-py:
     name: Build and test Python wrapper
     needs: [build-release]
+    permissions:
+      id-token: write  # IMPORTANT: this permission is mandatory for trusted publishing
 
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -221,15 +223,27 @@ jobs:
           - os: ubuntu-latest
             architecture: linux-aarch64
             plat-name: manylinux2014_aarch64
+            environment:
+              name: pypi
+              url: https://pypi.org/p/aries-askar
```
swcurran, line 228: This should not be aries-askar.  Ideally it is just “askar” — but lets find out if that is taken.  If it is, perhaps “owf-askar”?

jamshale: Oh. Changing the project and package name is a lot more work. aries-askar is in a tons of places. This was just trying to get a release published with the existing project.

swcurran:
OK — let’s leave it as that for now.

.github/workflows/build.yml
swcurran: Adding the comment again here so it is in the right place.  Sounds good — let’s leave the name as is for now.  We can talk about a change later.

swcurran (approved):

andrewwhitehead (approved):

PR #339, opened 2024-12-17
Adjust Pypi publish workflow

jamshale:
This is a new adjustment to the PyPi publish workflow. The existing method didn't work when trying to publish on mac and windows builds. 

Now it builds and uploads. Then downloads the artifacts separately after all the build steps complete and publishes all the packages together from ubuntu-latest os.


jamshale:
@swcurran @andrewwhitehead Tagging you here because I can't request reviewers.

andrewwhitehead (approved):

PR #354, opened 2025-01-28
Python wrapper entry cache adjustment

andrewwhitehead:
From the Python library documentation it appears that the native `lru_cache` decorator could result in caching of Entry instances longer than necessary. This PR switches to a custom cache implementation at the EntryList level which should avoid any undesired behaviour.

@ff137

### DIFF
--- a/wrappers/python/aries_askar/bindings/lib.py
+++ b/wrappers/python/aries_askar/bindings/lib.py
@@ -1,6 +1,7 @@
 """Library instance and allocated buffer handling."""
 
 import asyncio
+import functools
 import itertools
 import logging
 import os
@@ -46,6 +47,27 @@ LOG_LEVELS = {
     3: logging.INFO,
     4: logging.DEBUG,
 }
+
+
+def entry_cache(fn):
+    """Cache results for properties of individual entries."""
+
+    @functools.wraps(fn)
+    def wrapper(self, index: int):
+        if not hasattr(self, "_ecache"):
+            setattr(self, "_ecache", {})
+        cache = self._ecache
+        ckey = (fn, index)
+        if ckey in cache:
+            return cache[ckey]
+        res = fn(self, index)
+        cache[ckey] = res
+        if isinstance(res, dict):
+            # make sure the cached copy is not mutated
+            res = dict(res)
+        return res
+
+    return wrapper
 
 
 def _convert_log_level(level: Union[str, int, None]):
--- a/wrappers/python/aries_askar/bindings/lib.py
+++ b/wrappers/python/aries_askar/bindings/lib.py
@@ -8,6 +8,7 @@ import os
 import sys
 import threading
 import time
+from copy import deepcopy
 
 try:
     import orjson as json
@@ -59,12 +60,13 @@ def entry_cache(fn):
         cache = self._ecache
         ckey = (fn, index)
         if ckey in cache:
-            return cache[ckey]
-        res = fn(self, index)
-        cache[ckey] = res
+            res = cache[ckey]
+        else:
+            res = fn(self, index)
+            cache[ckey] = res
         if isinstance(res, dict):
             # make sure the cached copy is not mutated
-            res = dict(res)
+            res = deepcopy(res)
         return res
 
     return wrapper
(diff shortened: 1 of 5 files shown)


wrappers/python/aries_askar/bindings/lib.py
```diff
@@ -48,6 +49,27 @@
...
+        cache = self._ecache
+        ckey = (fn, index)
+        if ckey in cache:
+            return cache[ckey]
+        res = fn(self, index)
+        cache[ckey] = res
+        if isinstance(res, dict):
+            # make sure the cached copy is not mutated
+            res = dict(res)
+        return res
```
ff137, line 68: Is this correct? If res is a dict, it's cast to a dict?

ff137 (commented):
I think it's very much worthwhile to add a unit test that asserts caching works as expected.

GPT helped me generate the following -- maybe it's helpful inspiration, but obviously just auto-generated and requires more attention to detail:
```py
import unittest
from unittest.mock import patch, MagicMock
from aries_askar.bindings.handle import EntryListHandle

class TestEntryCache(unittest.TestCase):
    @patch('aries_askar.bindings.lib.Lib')
    def test_get_category_caching(self, mock_lib):
        # Setup the mock for Lib().invoke
        mock_invoke = MagicMock()
        # Define return values for different indices
        mock_invoke.side_effect = [
            'Category1',  # For index 0
            'Category2',  # For index 1
        ]
        mock_lib_instance = mock_lib.return_value
        mock_lib_instance.invoke.return_value = None  # Assuming invoke modifies the StrBuffer in place

        # Initialize EntryListHandle
        handle = EntryListHandle(value=123)

        with patch('aries_askar.bindings.lib.StrBuffer') as MockStrBuffer:
            # Mock StrBuffer to return the desired string
            mock_str_buffer_instance = MockStrBuffer.return_value
            mock_str_buffer_instance.__str__.return_value = mock_invoke()

            # First call with index 0
            category1_first = handle.get_category(0)
            self.assertEqual(category1_first, 'Category1')
            mock_lib_instance.invoke.assert_called_with(
                "askar_entry_list_get_category",
                (EntryListHandle, int, unittest.mock.ANY),
                handle,
                0,
                unittest.mock.ANY,
            )

            # Reset mock to track subsequent calls
            mock_lib_instance.invoke.reset_mock()

            # Second call with index 0 (should use cache)
            category1_second = handle.get_category(0)
            self.assertEqual(category1_second, 'Category1')
            mock_lib_instance.invoke.assert_not_called()

            # Call with a new index 1
            mock_str_buffer_instance.__str__.return_value = mock_invoke()  # Update return value
            category2 = handle.get_category(1)
            self.assertEqual(category2, 'Category2')
            mock_lib_instance.invoke.assert_called_with(
                "askar_entry_list_get_category",
                (EntryListHandle, int, unittest.mock.ANY),
                handle,
                1,
                unittest.mock.ANY,
            )

            # Reset mock to track subsequent calls
            mock_lib_instance.invoke.reset_mock()

            # Call again with index 1 (should use cache)
            category2_second = handle.get_category(1)
            self.assertEqual(category2_second, 'Category2')
            mock_lib_instance.invoke.assert_not_called()

if __name__ == '__main__':
    unittest.main()
```

wrappers/python/aries_askar/bindings/lib.py
andrewwhitehead: The `dict` constructor creates a new instance by iterating through the old instance's entries, but I changed it to a deep copy anyway in case there's nested structure.

andrewwhitehead:
I've added a unit test that should cover the `entry_cache` expectations, thanks.

andrewwhitehead:
Actually, there's a few more knock-on effects from renaming that would need some testing. I don't really see a need to rename the Python package, although we can rename the shared library that is produced by the build. The bundle produced by `build-xcframework.sh` produces a bundle with a hyperledger identifier which should be updated to `foundation.openwallet.askar`, although I don't know who would be affected by that change. Maybe we can get another release out first and then look at renaming the artifacts.

wrappers/python/aries_askar/bindings/lib.py
ff137: Aah I see now -- that modifying `res` downstream would modify the same object stored in cache. Was a lil confusing at first

ff137 (approved):
Nice 👍

swcurran (approved):

PR #373, opened 2025-03-18
Add support for prehashed signatures on EC curves

### LINKED ISSUE #371: Sign/verify with a different hashing function
Currently, SHA2-256 is matched with P256 (likely also with K256) and SHA2-384 is matched with P384.

Would it be possible to extend the verifying/signing interface to optionally provide a hashing algorithm to use that instead of the default one? And if none is supplied, just use the most common one (P256 -> SHA2-256, etc.).

We encountered this issue where an X.509 certificate was signed with P384, but the digest of the message was created using SHA2-256 and it took way to long to debug why Askar did not verify the signature correctly, but `@noble/curves` (which requires the user to hash before verifying) does verify it.

andrewwhitehead:
This adds support for three new (non-JWA) signing algorithms accepting pre-hashed input. The input may be larger or smaller than the field size, allowing mixing of (for example) a SHA-384 hash with a P-256 signing key.

Fixes #371

### DIFF
--- a/askar-crypto/src/alg/k256.rs
+++ b/askar-crypto/src/alg/k256.rs
@@ -2,7 +2,10 @@
 
 use k256::{
     ecdsa::{
-        signature::{Signer, Verifier},
+        signature::{
+            hazmat::{PrehashSigner, PrehashVerifier},
+            Signer, Verifier,
+        },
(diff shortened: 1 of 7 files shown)


TimoGlastra (approved):
this looks good to me!

PR #374, opened 2025-03-18
Add copying and renaming of individual Store profiles

### LINKED ISSUE #221: Ability to rename a profile and only export a specific profile
It would be useful if you can choose to export only a specific profile (or a subset of profiles) when calling `copyTo`.

In addition, I think it would also be nice if you can rename a profile. With Indy SDK import/export you could use a different wallet id after you had exported the wallet, and currently if you do export/import with Askar it will use the same profile name. 

I don't think there's a way currently to update a profile name after it has been created

andrewwhitehead:
Fixes #221

### DIFF
--- a/askar-storage/src/any.rs
+++ b/askar-storage/src/any.rs
@@ -65,6 +65,15 @@ impl<B: Backend> Backend for WrapBackend<B> {
     }
 
     #[inline]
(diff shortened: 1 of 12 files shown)


TimoGlastra (approved):
Thanks for the quick PR!!

PR #392, opened 2025-05-21
Update BLS key JWK encoding

andrewwhitehead:
This updates the JWK encoding for BLS12-381 keys to match `draft-ietf-cose-bls-key-representations-06`.

The combined `G1G2` key type is removed as it wouldn't make sense in the new representation. Instead, `G1` and `G2` keypairs can be converted directly, as long as the private key is present.


TimoGlastra (approved):
LGTM

PR #399, opened 2025-07-04
set page size for Android to 16kb

berendsliedrecht:
Signed-off-by: Berend Sliedrecht <berend@animo.id>


TimoGlastra:
@andrewwhitehead FYI this is needed for Android apps starting from November 2025: https://developer.android.com/guide/practices/page-sizes

PR #448, opened 2025-12-09
chore: update to 0.5

TimoGlastra:
Updates library version to 0.5, so we can release the new Askar features.


TimoGlastra:
@andrewwhitehead we might want to get some other changes in as well, like #449, and some of the dependabot updates?

PR #449, opened 2025-12-09
chore: remove secure env integration

### LINKED ISSUE #442: Remove `secure_env` integration
We have actually been using this directly from React Native/Credo rather than use the integration in Askar. For this reason I think it's better to remove the feature, as we're not really maintaining/using it.

cc @berendsliedrecht

TimoGlastra:
I kept some of the functionality for custom backends, but not sure what we want to keep?

I mainly wanted to remove the secure_env as we don't maintain it anymore, and then it's good to not keep it in here.

Fixes #442 

Probably good to wait until @berendsliedrecht is back, so he can also take a look

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,9 +39,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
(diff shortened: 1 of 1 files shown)


TimoGlastra (commented):
Thanks for fixing! Looks good to me

PR #470, opened 2026-06-24
fix: aarch64 TLS BadSignature from miscompiled ring assembly

### LINKED ISSUE #401: Database connectivity with ARM based arch
Connecting to agent-db on arm64 based arch fails with `invalid peer certificate: BadSignature`, when provided with the domain (as an endpoint) from the provider. Tested with RDS and Supabase, failed for both.

Also, in the same architecture if we provide combination of ip:port, it seems to connect correctly.  

We suspected some error with the provider setup. However, when tested to connect other services on arm based arch to the same db with similar config, it seems to be working just fine.

Just speculating if this could be some arch level dependency from some rust package, when resolving endpoints(which seems to be passing when there's no need to resolve the host).

Here's the error:
```sh
/app/node_modules/@credo-ts/askar/build/wallet/AskarWallet.js:183
            throw new core_1.WalletError(Error opening wallet ${[walletConfig.id](http://walletconfig.id/)}: ${error.message}, { cause: error });
                  ^
WalletError: Error opening wallet wallet-id: Error connecting to database pool
Caused by: error communicating with database: invalid peer certificate: BadSignature
    at AskarWallet._open (/app/node_modules/@credo-ts/askar/build/wallet/AskarWallet.js:183:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async AskarWallet.open (/app/node_modules/@credo-ts/askar/build/wallet/AskarWallet.js:120:9)
    at async WalletApi.open (/app/node_modules/@credo-ts/core/build/wallet/WalletApi.js:73:9)
    at async WalletApi.initialize (/app/node_modules/@credo-ts/core/build/wallet/WalletApi.js:47:13)
    at async Agent.initialize (/app/node_modules/@credo-ts/core/build/agent/BaseAgent.js:91:13)
    at async Agent.initialize (/app/node_modules/@credo-ts/core/build/agent/Agent.js:121:9)
    at async runRestAgent (/app/build/cliAgent.js:219:5)
    at async runCliServer (/app/build/cli.js:166:5) {
  [cause]: AriesAskarError: Error connecting to database pool
  Caused by: error communicating with database: invalid peer certificate: BadSignature
      at NodeJSAriesAskar.getAriesAskarError (/app/node_modules/@hyperledger/aries-askar-nodejs/build/NodeJSAriesAskar.js:90:16)
      at cb (/app/node_modules/@hyperledger/aries-askar-nodejs/build/NodeJSAriesAskar.js:41:44)
      at Object.<anonymous> (/app/node_modules/@[2060.io/ffi-napi/lib/callback.js:66:27](http://2060.io/ffi-napi/lib/callback.js:66:27)) {
    code: 1
  }
}
```

rblaine95:
## Problem

The published `linux-aarch64` library fails to open a TLS connection to Postgres (for example AWS RDS):

```
error communicating with database: invalid peer certificate: BadSignature
```

The same code and database work from `linux-x86_64`. A minimal Credo 0.7.0 reproduction is at https://github.com/rblaine95/askar-401-reproduction.

## Cause

This is the build toolchain, not the Rust code. The `linux-aarch64` and `linux-x86_64` libraries are cross-compiled in the `ghcr.io/rust-cross/manylinux2014-cross` images, which ship `binutils 2.29.1`. That assembler miscompiles the aarch64 assembly in `ring` (briansmith/ring#1728), which is why the workflow already sets `-D__ARM_ARCH=8`. The flag lets `ring` compile, but the emitted aarch64 crypto is wrong, so TLS signature verification fails. The x86_64 assembly is not affected, which is why only arm64 breaks.

Building the same source with a newer assembler produces a working library. Both images target the same glibc:

| image | gcc | binutils | glibc target |
| --- | --- | --- | --- |
| `rust-cross/manylinux2014-cross:aarch64` | 4.8.5 | 2.29.1 | 2.17 |
| `pypa/manylinux2014_aarch64` | 10.2.1 | 2.35 | 2.17 |

## Change

Build both Linux targets natively instead of cross-compiling:

- `linux-aarch64` on an `ubuntu-24.04-arm` runner
- `linux-x86_64` on `ubuntu-latest`

Each builds inside the matching `quay.io/pypa/manylinux2014_<arch>` image. Those images use `binutils 2.35` and still target `glibc 2.17`, so the libraries keep manylinux2014 compatibility. With a correct assembler the `-D__ARM_ARCH=8` workaround is no longer needed.

This removes `cross`, `Cross.toml`, and `CROSS_VERSION` from the Linux builds. The Android job still uses `cross` and is unchanged.

## Testing

I built the library from this branch in `pypa/manylinux2014_aarch64`, dropped it into `@openwallet-foundation/askar-nodejs`, and the Credo agent connected to an SSL RDS instance with no BadSignature error. The current release reproduces the error with the same setup. The built library reports `GLIBC_2.17` as its highest required symbol.

Closes #401

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,7 +3,6 @@ name: "Aries-Askar"
 env:
   RUST_VERSION: "1.86.0"
   RUST_NIGHTLY_VERSION: "nightly"
-  CROSS_VERSION: "0.2.4"
   TEST_FEATURES: "default"
 
 concurrency:
@@ -65,11 +64,6 @@ jobs:
       - name: Cargo check
         run: cargo check --workspace
 
-      - if: ${{ runner.os == 'Linux' }}
-        name: Pre-install cross
-        run: |
-          cargo install --bins --git https://github.com/rust-embedded/cross --locked --tag v${{ env.CROSS_VERSION }} cross
-
   tests:
     name: Run tests
     needs: [checks]
@@ -123,15 +117,15 @@ jobs:
       matrix:
         include:
           - architecture: linux-aarch64
-            os: ubuntu-latest
+            os: ubuntu-24.04-arm
             lib: libaries_askar.so
             target: aarch64-unknown-linux-gnu
-            use_cross: true
+            manylinux_image: quay.io/pypa/manylinux2014_aarch64
           - architecture: linux-x86_64
             os: ubuntu-latest
             lib: libaries_askar.so
             target: x86_64-unknown-linux-gnu
-            use_cross: true
+            manylinux_image: quay.io/pypa/manylinux2014_x86_64
           - architecture: darwin-universal
             os: macos-latest
             lib: libaries_askar.dylib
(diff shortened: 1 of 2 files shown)


andrewwhitehead (approved):
Thanks for debugging that!

